PR #5, opened 2023-08-31
Bmtagger

KateSakharova:
Added bmtagger tool (locally everything works with conda!), singularity test crashes by some reason, maybe memory?


modules/ebi-metagenomics/bmtagger/index_reference/main.nf
```diff
@@ -0,0 +1,32 @@
+process CREATE_DB_BMTAGGER {
```
mberacochea, line 1: This one is missing the versions.yml output file.

modules/ebi-metagenomics/bmtagger/index_reference/main.nf
```diff
@@ -0,0 +1,32 @@
...
+
+    output:
+    path("${reference_fasta.baseName}.bitmask")         , emit: bitmask
+    path("${reference_fasta.baseName}.srprism.*")       , emit: srprism
+    path("${reference_fasta}.n*")              , emit: blast_db
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
```
mberacochea, line 21: I think we should include the args too.. for example to tweak the parameters of the different tools (such as -A and -w"

mberacochea (changes requested):
Nice :), added 2 minor comments and also there are 2 failling CI jobs.

modules/ebi-metagenomics/bmtagger/index_reference/main.nf
mberacochea: Fixed

PR #12, opened 2023-09-12
Eggnogmapper

vagkaratzas:


modules/ebi-metagenomics/eggnogmapper/main.nf
```diff
@@ -0,0 +1,56 @@
...
+        --data_dir ${eggnog_data_dir} \\
+        -m diamond \\
+        --dmnd_db ${eggnog_diamond_db} \\
+        --database ${eggnog_db} \\
+        --output ${prefix} \\
+        $args
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        eggnog-mapper: $VERSION
```
mberacochea, line 40: It's better to capture the version from the emapper.py script directly:
```
\$(echo \$(emapper.py --version) | grep -o "emapper-[0-9]\\+\\.[0-9]\\+\\.[0-9]\\+" | sed "s/emapper-//")
```
this expression should do the trick

tests/modules/ebi-metagenomics/eggnogmapper/test.yml
```diff
@@ -0,0 +1,8 @@
+- name: eggnogmapper test_eggnogmapper
+  command: nextflow run ./tests/modules/ebi-metagenomics/eggnogmapper -entry test_eggnogmapper -c ./tests/config/nextflow.config -c ./tests/modules/ebi-metagenomics/eggnogmapper/nextflow.config
+  tags:
+    - eggnogmapper
+  files:
+    - path: output/eggnogmapper/test.emapper.hits
+      md5sum: f8629819c48c888275dbb1a09749c891
+    - path: output/eggnogmapper/versions.yml
```
mberacochea, line 8: You are missing the md5sum of the `versions.yml` file

modules/ebi-metagenomics/eggnogmapper/main.nf
```diff
@@ -0,0 +1,56 @@
...
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def VERSION = '2.1.12'
+    """
+    emapper.py \\
+        --cpu ${task.cpus} \\
+        -i ${fasta} \\
```
mberacochea, line 30: Add support got .gz fasta files please

mberacochea (changes requested):
I've added some comments. Thank you @vagkaratzas, this is great stuff

modules/ebi-metagenomics/eggnogmapper/main.nf
vagkaratzas: In the meta.yml file the fasta pattern is: pattern: "*.{fasta,fa,fasta.gz,fa.gz}"
Should work fine, right?

mberacochea (approved):
:shipit:

PR #18, opened 2023-10-05
Eggnogmapper

vagkaratzas:


modules/ebi-metagenomics/diamond/makedb/main.nf
```diff
@@ -0,0 +1,35 @@
...
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    """
+    diamond \\
+        makedb \\
+        --threads $task.cpus \\
+        --in  $fasta \\
```
mberacochea, line 26:
```suggestion
        --in $fasta \\
```

modules/ebi-metagenomics/eggnogmapper/main.nf
```diff
@@ -25,6 +25,7 @@ process EGGNOGMAPPER {
     def prefix = task.ext.prefix ?: "${meta.id}"
     def is_compressed = fasta.name.endsWith(".gz")
     def fasta_name = fasta.name.replace(".gz", "")
+    def dbmem = task.memory.toMega() > 40000 ? '--dbmem' : ''
```
mberacochea, line 28: Should we have this parameter added explicitly instead of linking it with the memory requirements?. The DB alone is like 40G

mberacochea (commented):
The only bit I'm not sure is the dbmem parameter

modules/ebi-metagenomics/eggnogmapper/main.nf
vagkaratzas: This is a failsafe for different profiles. For example, eggnog requires >40GB to load the db in memory, which my local profile for laptop cannot provide and therefore never finishes execution. Couldn't find a better place for this check

mberacochea: Yeah, I understand but I'm not sure if "forcing" the db load when asking for loads memory is the best. I generaly think that is what our pipelines would like to do.
I think it's safe to leave it... we can change it if needed

mberacochea (approved):
:shipit:

PR #19, opened 2023-10-06
Deeptmhmm

vagkaratzas:


modules/ebi-metagenomics/deeptmhmm/meta.yml
```diff
@@ -0,0 +1,45 @@
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/yaml-schema.json
+name: deeptmhmm
+description: A Deep Learning Model for Transmembrane Topology Prediction and Classification
+keywords:
+  - sort
```
mberacochea, line 6: sort, example? I think you forgot the keyworks

modules/ebi-metagenomics/deeptmhmm/meta.yml
```diff
@@ -0,0 +1,45 @@
...
+  - fasta:
+      type: file
+      description: Database of sequences in FASTA format
+      pattern: "*.{fasta,fa,fasta.gz,fa.gz}"
+
+output:
+  - gff3:
+      type: file
+      description: Predicted topologies in general Feature Format Version 3
+      pattern: "biolib_results/TMRs.gff3"
```
mberacochea, line 27: Is this correct? According to the docs it should be:

- input/output patterns (if present) MUST follow a [Java glob pattern](https://docs.oracle.com/javase/tutorial/essential/io/fileOps.html#glob)

mberacochea: should only be for the file? and not the path?

mberacochea (changes requested):
Great stuff... keep them coming. I've added some comments

modules/ebi-metagenomics/deeptmhmm/meta.yml
vagkaratzas: I blame the linting. Should have reminded me :P

modules/ebi-metagenomics/deeptmhmm/meta.yml
vagkaratzas: Nah this is more like it. I updated it with some more info and also added the optional outputs I had forgotten.

mberacochea (approved):
:shipit:

PR #21, opened 2023-10-23
Add input parameter (save_merged) to decide whether to merge paired-e…

chrisAta:
…nd files or not, as if you do the _1 and _2 output files only contain reads that couldn't be merged rather than all cleaned _1 and _2 reads

### DIFF
--- a/tests/modules/ebi-metagenomics/fastp/main.nf
+++ b/tests/modules/ebi-metagenomics/fastp/main.nf
@@ -15,6 +15,6 @@ workflow test_fastp {
     FASTP (
         input,
         params.save_trimmed_fail,
-        params.save_merged
+        true
      )
 }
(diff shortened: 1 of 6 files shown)


vagkaratzas (approved):

tests/modules/ebi-metagenomics/fastp/main.nf
```diff
@@ -15,6 +15,6 @@ workflow test_fastp {
     FASTP (
         input,
         params.save_trimmed_fail,
-        params.save_merged
+        true
```
mberacochea, line 18:
```suggestion
        true // saved merged reads
```

mberacochea (approved):
Nice, thank you @chrisAta . I've left one very minor comment, but other than that :shipit:

PR #22, opened 2023-10-30
Adding Mapseq module

chrisAta:
Still have to add mapseq2biom.py and krona modules to make the mapseq_otu_krona subworkflow, but these will come once we have our new bioconda repo. To not make the code review super complicated I'll do a pull request one module at a time


KateSakharova (approved):
Looks very nice to me! Thank you, Chris!

modules/ebi-metagenomics/mapseq/meta.yml
```diff
@@ -0,0 +1,56 @@
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/yaml-schema.json
+name: "mapseq"
+description: Perform taxonomic classification of rRNA reads using reference databases
+keywords:
+  - taxonomy
+  - reads
+  - otu
```
mberacochea, line 8: shouldn't be `OTU`?

modules/ebi-metagenomics/mapseq/meta.yml
```diff
@@ -0,0 +1,56 @@
...
+  - rRNA
+tools:
+  - "mapseq":
+      description: "Open source metagenomic 16S/18S read classifier enabling comparative metagenomics."
+      documentation: https://github.com/jfmrod/MAPseq
+      doi: "10.1093/bioinformatics/btx517"
+      licence: "['dual-licensed under GPL v3+ or BSD 2-clause']"
```
mberacochea, line 15: I can only see GPL 3 in the licence -> https://github.com/jfmrod/MAPseq/blob/master/COPYING

mberacochea (approved):
Good stuff @chrisAta. I've added some minor comments, but other than that it looks good to me.

modules/ebi-metagenomics/mapseq/meta.yml
chrisAta: I thought keywords were always lower-case? I found some examples of `OTU` being written in lower case in a few nf-core modules (though admittedly written by a lot of the same people):

https://github.com/nf-core/modules/blob/master/modules/nf-core/taxpasta/standardise/meta.yml
https://github.com/nf-core/modules/blob/master/modules/nf-core/taxpasta/merge/meta.yml
https://github.com/nf-core/modules/blob/master/modules/nf-core/motus/merge/meta.yml

modules/ebi-metagenomics/mapseq/meta.yml
chrisAta: huh interesting - this field was filled automatically from the bioconda repo: https://anaconda.org/bioconda/mapseq

The COPYING file in the github repo was last updated 6 years ago, maybe it's just outdated?

mberacochea: ahh.. then I would just leave it. The code uses another tool which has a different licence, I couldn't find it... so I guess that is the reason for the mixed thingy.

modules/ebi-metagenomics/mapseq/meta.yml
mberacochea: then... don't pay attention to me. I'm a big fan of lowercasing everything

PR #24, opened 2023-11-10
esmfold CPU only

vagkaratzas:
Won't be able to be pushed directly to nf-core in its current state.
We will need to either upload the current cpu-only working conda environment publicly,
or create and upload a working GPU/CPU conda environment.


vagkaratzas:
If want to merge into main, we will need to currently remove checks for docker and singularity images. Else we keep it in a separate branch.

mberacochea:
@vagkaratzas is this one still relevant?

vagkaratzas:
> @vagkaratzas is this one still relevant?

Will never pass tests because one will need to copy my conda environment, local or codon one to their machine. Other than that, it would be useful for anyone that wants to run protein structure predictions. It also has a GPU (codon env) mode now

mberacochea:
I close this one as it was running the actions everytime main was updated... wasting GA credits

PR #26, opened 2023-12-14
Esl sfetch

chrisAta:


vagkaratzas (commented):
Minor comment. When using more than one input meta, give them different names (meta1, meta2), because it is unsure which of the two will be use in naming prefix and files later on. Add both to meta.yml

modules/ebi-metagenomics/easel/eslsfetch/main.nf
```diff
@@ -0,0 +1,69 @@
...
+        'https://depot.galaxyproject.org/singularity/easel:0.49--h031d066_0':
+        'biocontainers/easel:0.49--h031d066_0' }"
+
+    input:
+    tuple val(meta), path(fasta)
+    tuple val(meta), path(cmsearch_deoverlap_out)
+
+    output:
+    tuple val(meta), path("*easel_coords.fasta"), emit: easel_coords
+    tuple val(meta), path("*matched_seqs_with_coords"), emit: matched_seqs_with_coords
```
vagkaratzas, line 17: tuple val(meta), path("*easel_coords.fasta")              , emit: easel_coords
tuple val(meta), path("*matched_seqs_with_coords"), emit: matched_seqs_with_coords

chrisAta:
> 
makes sense, though in this case the meta will be the exact same for both honestly.. maybe in that case I only leave the meta for the first input and remove it for the second?

modules/ebi-metagenomics/easel/eslsfetch/main.nf
```diff
@@ -0,0 +1,69 @@
...
+    if [ "$is_compressed" == "true" ]; then
+        gzip -c -d $fasta > $fasta_name
+    fi
+
+    awk \\
+        '{print \$1"-"\$3"/"\$8"-"\$9" "\$8" "\$9" "\$1}' \\
+        $cmsearch_deoverlap_out \\
+        > ${prefix}.matched_seqs_with_coords
+
+    esl-sfetch \\
```
KateSakharova, line 41: IMHO we need to cut that module into 2 with index and without.. 
and _ideally_ check is there index already exist - do not index again

modules/ebi-metagenomics/easel/eslsfetch/main.nf
```diff
@@ -0,0 +1,69 @@
...
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+
+    def is_compressed = fasta.name.endsWith(".gz")
+    def fasta_name= fasta.name.replace(".gz", "")
+
```
vagkaratzas, line 29: Switch to this, more Groovy and only checks extension as should:
def is_compressed = fasta.getExtension() == "gz" ? true : false
def fasta_name = is_compressed ? fasta.getBaseName() : fasta

modules/ebi-metagenomics/easel/eslsfetch/main.nf
chrisAta: I think it's something we can easily change in the future, but for now we think it isn't worth it due to overhead of creating more than 1 job for this module when it's a pretty fast process anyway

vagkaratzas:
> > Minor comment. When using more than one input meta, give them different names (meta1, meta2), because it is unsure which of the two will be use in naming prefix and files later on. Add both to meta.yml
> 
> makes sense, though in this case the meta will be the exact same for both honestly.. maybe in that case I only leave the meta for the first input and remove it for the second?

Aye!

modules/ebi-metagenomics/easel/eslsfetch/meta.yml
```diff
@@ -0,0 +1,49 @@
...
+        Groovy Map containing sample information
+        e.g. `[ id:'test', single_end:false ]`
+  - fasta:
+      type: file
+      description: Fasta sequence file from amplicon sequencing
+      pattern: "*.{fasta, fasta.gz, fa, fa.gz}"
+  - cmsearch_deoverlap_out:
+      type: file
+      description: Tabular output from 'cmsearchtnloutdeoverlap' module
+      pattern: "*.{tblout.deoverlap}"
```
KateSakharova, line 29: maybe that name should not be hardcoded here `tblout.deoverlap`
probably it is just enough saying it should be tab separated file

KateSakharova (approved):

modules/ebi-metagenomics/easel/eslsfetch/meta.yml
vagkaratzas: Agree with Kate, just state the extension.
Same here:
*easel_coords.fasta -> *.fasta

chrisAta: > Agree with Kate, just state the extension. Same here: *easel_coords.fasta -> *.fasta

I specify this one because the input is also a fasta file and I don't want it to grab both as output

modules/ebi-metagenomics/easel/eslsfetch/main.nf
mberacochea: No need for the `? true : false` as `fasta.getExtension() == "gz"` boolean expression.

modules/ebi-metagenomics/easel/eslsfetch/main.nf
```diff
@@ -0,0 +1,68 @@
...
+    esl-sfetch \\
+        -Cf \\
+        $fasta_name \\
+        ${prefix}.matched_seqs_with_coords \\
+        > ${prefix}_easel_coords.fasta
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        : \$(esl-sfetch -h | grep -o '^# Easel [0-9.]*' | sed 's/^# Easel *//')
```
mberacochea, line 52: easl is missing from the versions.yml thing

mberacochea (commented):
I'm a bit late.. but I added 2 comments

PR #29, opened 2023-12-15
Swf/rrna extraction

chrisAta:
rRNA extraction subworkflow to use in our pipelines


subworkflows/ebi-metagenomics/rrna_extraction/main.nf
```diff
@@ -0,0 +1,42 @@
...
+    take:
+    ch_fasta // channel: [ val(meta), [ fasta ] ]
+
+    main:
+
+    ch_versions = Channel.empty()
+
+    INFERNAL_CMSEARCH(
+        ch_fasta,
+        file(params.rfam)
```
KateSakharova, line 17: Is rfam a file or folder with models? (probably it would be better to add that to name like `rfam_model_file`/`rfam_model_dir`)
I'm wondering how good is to take db directly from params. We can re-use the same subwf in one pipeline with various sets of rfams..

subworkflows/ebi-metagenomics/rrna_extraction/main.nf
```diff
@@ -0,0 +1,42 @@
...
+    )
+    ch_versions = ch_versions.mix(INFERNAL_CMSEARCH.out.versions.first())
+
+    CMSEARCHTBLOUTDEOVERLAP(
+        INFERNAL_CMSEARCH.out.cmsearch_tbl,
+        file(params.rfam_clan)
```
KateSakharova, line 23: same here.. `rfam_clan_file`

subworkflows/ebi-metagenomics/rrna_extraction/main.nf
```diff
@@ -0,0 +1,42 @@
...
+include { INFERNAL_CMSEARCH           } from '../../../modules/ebi-metagenomics/infernal/cmsearch/main'
+include { CMSEARCHTBLOUTDEOVERLAP     } from '../../../modules/ebi-metagenomics/cmsearchtbloutdeoverlap/main'
+include { EASEL_ESLSFETCH             } from '../../../modules/ebi-metagenomics/easel/eslsfetch/main'
+
+workflow RRNA_EXTRACTION {
+
+    take:
+    ch_fasta // channel: [ val(meta), [ fasta ] ]
+
+    main:
```
KateSakharova, line 11: In motus-pipeline we have chunking fasta file before cmsearch. Should we think about it here as well? Or will subwf work if I get chunked fasta as input (I guess no, as deoverlap requires a full/merged output from cmsearch).
It it just a thought.. I don't know would it work fast on huge fastas.

tests/subworkflows/ebi-metagenomics/rrna_extraction/nextflow.config
```diff
@@ -0,0 +1,14 @@
...
+    publishDir = { "${params.outdir}/${task.process.tokenize(':')[-1].tokenize('_')[0].toLowerCase()}" }
+    
+    withName: INFERNAL_CMSEARCH {
+    ext.args = '--cut_ga'
+    }
+}
+
+params {
+
+    rfam = "/hps/nobackup/rdf/metagenomics/service-team/users/chrisata/scripts_taxdb_nf_test/RFAM/14.10/ribo.cm"
```
KateSakharova, line 12: Does it work on github actions? Probably you can add heavy files on our FTP and pull from there?

KateSakharova (changes requested):
Nice! I've added some comments

subworkflows/ebi-metagenomics/rrna_extraction/main.nf
chrisAta: rfam is a file, but good idea for folders future modules!

I think we can still take it from params even if it's used more than once because you could just add another param for each rfam/database.. do you have a suggestion for a better way though?

subworkflows/ebi-metagenomics/rrna_extraction/main.nf
chrisAta: good point - we would need to chunk before cmsearch and then merge them again before deoverlap..especially because cmsearch can be so slow on big files, definitely something to think about

tests/subworkflows/ebi-metagenomics/rrna_extraction/nextflow.config
chrisAta: oh good point I forgot about this in the test.. I'll replace it with a small chunk of our Rfam file for now

chrisAta: but long term I agree would be good to have these on the FTP

subworkflows/ebi-metagenomics/rrna_extraction/main.nf
chrisAta: done!

tests/subworkflows/ebi-metagenomics/rrna_extraction/nextflow.config
chrisAta: fixed this by adding small chunk of rfam as test files

subworkflows/ebi-metagenomics/rrna_extraction/main.nf
```diff
@@ -0,0 +1,44 @@
...
+    val_rfam     // value: /path/to/rfam for cmsearch
+    val_claninfo // value: /path/to/claninfo for cmsearchtbloutdeoverlap
+
+    main:
+
+    ch_versions = Channel.empty()
+
+    INFERNAL_CMSEARCH(
+        ch_fasta,
+        file(val_rfam)
```
mberacochea, line 19: Do we need to convert them to "file"? Why don't we just ask users to send a "file"?. If would be easier to catch if the file doesn't exist for example

mberacochea (changes requested):
I think this one is in pretty good shape. I left one minor comment

subworkflows/ebi-metagenomics/rrna_extraction/main.nf
chrisAta: yes that makes sense! I'll change this and convert them into files in the unit test as a good example

subworkflows/ebi-metagenomics/rrna_extraction/main.nf
```diff
@@ -0,0 +1,44 @@
...
+
+    INFERNAL_CMSEARCH(
+        ch_fasta,
+        val_rfam
+    )
+    ch_versions = ch_versions.mix(INFERNAL_CMSEARCH.out.versions.first())
+
+    CMSEARCHTBLOUTDEOVERLAP(
+        INFERNAL_CMSEARCH.out.cmsearch_tbl,
+        val_claninfo
```
mberacochea, line 25:
```suggestion
        claninfo
```

mberacochea (commented):
Thanks @chrisAta , I just renamed val_rfam and val_claninfo

subworkflows/ebi-metagenomics/rrna_extraction/main.nf
chrisAta: all done thank you!

mberacochea (approved):
Excellent, thank you @chrisAta

KateSakharova (approved):

PR #31, opened 2024-01-16
Cutadapt

chrisAta:
There already exists a cutadapt nf-core module (https://github.com/nf-core/modules/blob/master/modules/nf-core/cutadapt/main.nf), but for the amplicon pipeline we need it to dynamically be able to change not just single/paired mode, but also which strands should get trimmed of primers (either both, just forward, or just reverse), so I copied the base of the module but also made a couple changes that will implement this behaviour


modules/ebi-metagenomics/cutadapt/meta.yml
```diff
@@ -0,0 +1,62 @@
...
+      type: file
+      description: |
+        List of at most two primer fasta files
+        The first file should be the forward strand
+        The second file should be the reverse strand
+        If you only want to trim a primer from one strand,
+        then use an empty file as a standin for the second
+        e.g. `[ path/to/fwd_primer, empty.txt ]` for only using
+        the forward priemer
+      pattern: "*.{fasta}"
```
vagkaratzas, line 40: if it's only one option, you can remove the brackets:
pattern: "*.fasta"

vagkaratzas (approved):
Very clever way to add the primer files. I am not very certain about the various conditions for rev_primer, but I think you would capture anything wrong with the tests.

modules/ebi-metagenomics/cutadapt/main.nf
```diff
@@ -0,0 +1,56 @@
...
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def trimmed  = meta.single_end ? "-o ${prefix}.trim.fastq.gz" : "-o ${prefix}_1.trim.fastq.gz -p ${prefix}_2.trim.fastq.gz"
+    def fwd_primer = primers[0].size() > 0 ? "-g file:${primers[0]}"  : ''
+    def rev_primer = primers[1].size() > 0  && meta.single_end ? "-a file:${primers[1]}" : primers[1].size() > 0  && !meta.single_end ? "-G file:${primers[1]}" : ''
```
mberacochea, line 27: I would split this as conventional "if" statements.. it's a bit hard to read IMO

mberacochea (changes requested):
Good stuff.. I left one minor comment

modules/ebi-metagenomics/cutadapt/main.nf
```diff
@@ -0,0 +1,56 @@
+
+process CUTADAPT {
+    tag "$meta.id"
+    label 'process_single'
```
vagkaratzas, line 4: label 'process_medium'
in nf-core. Make sure you got the preferred one here.

modules/ebi-metagenomics/cutadapt/main.nf
```diff
@@ -0,0 +1,56 @@
...
+
+    conda "bioconda::cutadapt=4.6"
```
vagkaratzas, line 6: Also, in nf-core they've stopped writing this and instead do this:
conda "${moduleDir}/environment.yml"
And the resepctive file:
name: cutadapt
channels:
  - conda-forge
  - bioconda
  - defaults
dependencies:
  - bioconda::cutadapt=3.4

KateSakharova (approved):
Looks good to me, just resolve complains from nextflow-preacher :)

chrisAta:
Thanks everyone for the great feedback! I implemented all the changes now so I think it's ready 🚀

mberacochea (approved):
:shipit:

PR #33, opened 2024-03-04
Bug/decont bwamem2

Ales-ibt:
For decontamination it makes no sense to generate ${prefix}_other.fq.gz and ${prefix}_singleton.fq.gz files for paired-end library layout. Test was updated accordingly.

### DIFF
--- a/modules/ebi-metagenomics/samtools/bam2fq/main.nf
+++ b/modules/ebi-metagenomics/samtools/bam2fq/main.nf
@@ -29,8 +29,8 @@ process SAMTOOLS_BAM2FQ {
             -@ $task.cpus \\
             -1 ${prefix}_1.fq.gz \\
             -2 ${prefix}_2.fq.gz \\
-            -0 ${prefix}_other.fq.gz \\
-            -s ${prefix}_singleton.fq.gz \\
+            -0 /dev/null \\
+            -s /dev/null \\
             $inputbam
 
         cat <<-END_VERSIONS > versions.yml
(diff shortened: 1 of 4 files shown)


modules/ebi-metagenomics/samtools/bam2fq/main.nf
```diff
@@ -29,8 +29,8 @@ process SAMTOOLS_BAM2FQ {
             -@ $task.cpus \\
             -1 ${prefix}_1.fq.gz \\
             -2 ${prefix}_2.fq.gz \\
-            -0 ${prefix}_other.fq.gz \\
-            -s ${prefix}_singleton.fq.gz \\
+            -0 /dev/null \\
```
mberacochea, line 32: I think we should keep those files, someone may need them. I propose we split the outputs:
```
    output:
    tuple val(meta), path("*{_1,_2}.fq.gz"),      optional: true, emit: reads
    tuple val(meta), path("*_interleaved.fq.gz"), optional: true, emit: interleaved_reads
    tuple val(meta), path("*_singleton.fq.gz"),   optional: true, emit: singleton_reads
    tuple val(meta), path("*_other.fq.gz"),       optional: true, emit: other_reads
    path "versions.yml",                                          emit: versions
```
or even merging _1,_2 and _interleaved:
```
    output:
    tuple val(meta), path("*{_1,_2,_interleaved}.fq.gz"),         emit: reads
    tuple val(meta), path("*_singleton.fq.gz"),   optional: true, emit: singleton_reads
    tuple val(meta), path("*_other.fq.gz"),       optional: true, emit: other_reads
    path "versions.yml",                                          emit: versions
```

mberacochea (changes requested):

mberacochea (approved):

PR #34, opened 2024-04-09
New module/bwamem2 decont nobams

Ales-ibt:
New decontamination module that generates decont fastq files on the fly without bam outputs.

### DIFF
--- a/modules/ebi-metagenomics/bwamem2decontnobams/main.nf
+++ b/modules/ebi-metagenomics/bwamem2decontnobams/main.nf
@@ -5,7 +5,7 @@ process BWAMEM2DECONTNOBAMS {
     conda "${moduleDir}/environment.yml"
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
         'https://depot.galaxyproject.org/singularity/bwa-mem2:2.2.1--hd03093a_5':
-        'quay.io/biocontainers/bwa-mem2:2.2.1--hd03093a_5' }"
+        'biocontainers/bwa-mem2:2.2.1--hd03093a_5' }"
 
     input:
     tuple val(meta), path(reads)
--- a/modules/ebi-metagenomics/bwamem2decontnobams/main.nf
+++ b/modules/ebi-metagenomics/bwamem2decontnobams/main.nf
@@ -1,0 +1,54 @@
+process BWAMEM2DECONTNOBAMS {
+    tag "$meta.id"
+    label 'process_high'
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/bwa-mem2:2.2.1--hd03093a_5':
+        'quay.io/biocontainers/bwa-mem2:2.2.1--hd03093a_5' }"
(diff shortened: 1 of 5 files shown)


modules/ebi-metagenomics/bwamem2decontnobams/main.nf
```diff
@@ -0,0 +1,55 @@
...
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/mulled-v2-e5d375990341c5aef3c9aff74f96f66f65375ef6:2d15960ccea84e249a150b7f5d4db3a42fc2d6c3-0' :
+        'biocontainers/mulled-v2-e5d375990341c5aef3c9aff74f96f66f65375ef6:2d15960ccea84e249a150b7f5d4db3a42fc2d6c3-0' }"
+
+
+    input:
+    tuple val(meta), path(reads)
+    tuple val(meta2), path(index)
```
chrisAta, line 13: This isn't a request for a change, you can keep it this way but just starting a conversation about something...

`meta2` Is this the standard for when you need more than one meta? I think the official definition of `meta` in nf-core is metadata about samples, and `meta2` here is the reference genome.. so I wonder whether it makes more sense to be more explicit when we include more than one meta-type input

modules/ebi-metagenomics/bwamem2decontnobams/meta.yml
```diff
@@ -0,0 +1,56 @@
...
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. `[ id:'sample1', single_end:false ]`
+  - reads:
+      type: file
+      description: |
+        List of input FastQ files of size 1 and 2
+        for single-end and paired-end data, respectively
+  - meta2:
```
chrisAta, line 28: Just pointing out `meta2` here with regards to my last comment

chrisAta (approved):
Looks great to me thanks! Left a comment to start a little conversation about something we might want to consider as a team, but I don't really have any changes I think should be made

modules/ebi-metagenomics/bwamem2decontnobams/main.nf
KateSakharova: Do we need a meta here at all? If that is index of reference genome we know all info from filename and location (name, version, organism, etc). What do we store in meta2?

Ales-ibt: Yes, the idea here is to track the different reference genomes we are using in each round of decontamination. There's a discussion about using 'meta' and 'meta2', and this form is the standard at the moment for [Best practices for multiple process inputs](https://github.com/nf-core/modules/issues/4311)

Ales-ibt:
Thank you for your review. Merging branches now 🚀

modules/ebi-metagenomics/bwamem2decontnobams/main.nf
chrisAta: > There's a discussion about using 'meta' and 'meta2', and this form is the standard at the moment for [Best practices for multiple process inputs](https://github.com/nf-core/modules/issues/4311)

sounds good to me then, keep it as is for sure!

PR #35, opened 2024-04-09
Add meta2.is as an extra prefix to outputs

Ales-ibt:
I have found the answer to why to use meta2.id in this module. I am using it now to name the outputs, otherwise, the name of input and output files is the same when more than one round of decontamination is needed, which makes the process crash.


KateSakharova (approved):
Nice, thank you!

chrisAta (approved):
Looks good to me!

PR #37, opened 2024-04-17
New module/krona/ktimporttext (mostly copied from main nf-core modules repo)

chrisAta:
This is mostly a copy of the existing nf-core krona/ktimporttext module (https://nf-co.re/modules/krona_ktimporttext) so that we can add it to subworkflows. The only changes I made are the following:

- Added a `stub` to the module
- Converted the unit tests from `Pytest` to `nf-test`
- Updated `meta.yml` based on more updated nf-core linting requirements
- Added myself as an author alongside the existing author, and just myself as maintainer


KateSakharova (approved):
Difficult to review because we don't see initial subwfs :D 
Looks good!

PR #41, opened 2024-05-09
Add module/extractcoords

chrisAta:
Add extractcoords module that uses mgnify-pipelines-toolkit. Will then be added to the rrna_extraction subworkflow


modules/ebi-metagenomics/extractcoords/meta.yml
```diff
@@ -0,0 +1,58 @@
...
+        Groovy Map containing sample information
+        e.g. `[ id:'sample1', single_end:false ]`
+  - easel_coords_fasta:
+      type: file
+      description: Fasta file output from running esl-sfetch to extract sequences by name
+      pattern: "*.fasta"
+  - matched_seqs_with_coords:
+      type: file
+      description: cmsearchdeoverlap output formatted for use by esl-sfetch
+      pattern: "*.matched_seqs_with_coords"
```
vagkaratzas, line 31: Is this the actual output format of esl-esfetch? Anything more standardised?

modules/ebi-metagenomics/extractcoords/meta.yml
```diff
@@ -0,0 +1,58 @@
...
+      description: Fasta file containing the SSU sequences
+      pattern: "*.fasta"
+  - lsu_fasta:
+      type: file
+      description: Fasta file containing the LSU sequences
+      pattern: "*.fasta"
+  - concat_ssu_lsu_coords:
+      type: file
+      description: Concatenated file of coordinates of matches of both SSU and LSU
+      pattern: "*.concat_SSU_LSU_coords"
```
vagkaratzas, line 50: Same here. Depending on the type I would rename to {meta}.tsv or something

vagkaratzas: Or worst case scenario {meta}.concat_SSU_LSU_coords.tsv. But I would still go for something shorter ;)

modules/ebi-metagenomics/extractcoords/meta.yml
chrisAta: It's a space separated file, not tab or comma.. I guess I could use .txt in that case?

modules/ebi-metagenomics/extractcoords/meta.yml
chrisAta: same here, it's space separated, but .txt file is probably better

modules/ebi-metagenomics/extractcoords/main.nf
```diff
@@ -0,0 +1,51 @@
...
+        'biocontainers/mgnify-pipelines-toolkit:0.1.2--pyhdfd78af_0' }"
+
+    input:
+    tuple val(meta), path(easel_coords_fasta)
+    tuple val(meta), path(matched_seqs_with_coords)
+
+    output:
+    tuple val(meta),  path("sequence-categorisation/*SSU.fasta") , optional: true, emit: ssu_fasta
+    tuple val(meta),  path("sequence-categorisation/*LSU.fasta") , optional: true, emit: lsu_fasta
+    tuple val(meta),  path("*concat_SSU_LSU_coords.txt")             , emit: concat_ssu_lsu_coords
```
vagkaratzas, line 18: remove spaces before comma because of rename

vagkaratzas (approved):

Ales-ibt (approved):
Thanks a lot. It looks ok to me :D

PR #42, opened 2024-05-13
Rrna extraction swf/add extractcoords

chrisAta:
Add new extractcoords module to rrna_extraction subworkflow, including new nf-test unit tests


vagkaratzas (approved):
LGTM

PR #44, opened 2024-06-11
Antismash module

chrisAta:


modules/ebi-metagenomics/antismash/main.nf
```diff
@@ -0,0 +1,67 @@
...
+    path "versions.yml"                                         , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def is_compressed = antismash_db.name.endsWith(".tar.gz")
+    def antismashdb_name = antismash_db.name.replace(".tar.gz", "")
```
mberacochea, line 27: Should we support a compressed `antismash_db`? For our pipeline it will introduce an significant overhead.

modules/ebi-metagenomics/antismash/main.nf
```diff
@@ -0,0 +1,67 @@
...
+        tar -xzvf $antismash_db
+    fi
+
+    antismash \\
+        -c ${task.cpus} \\
+        ${args} \\
+        --databases ${antismashdb_name} \\
+        --genefinding-gff3 ${genes} \\
+        --output-dir ${prefix}_results \\
+        ${contigs}
```
mberacochea, line 40: We should supprot gbk inputs too. For example -> https://github.com/EBI-Metagenomics/mettannotator/blob/dev/modules/local/antismash.nf uses a gbk instead of a gff.

Or should we modify the mettannotator one instead? I'm not sure.

mberacochea (changes requested):

modules/ebi-metagenomics/antismash/main.nf
chrisAta: makes sense, I will add that

modules/ebi-metagenomics/antismash/main.nf
chrisAta: I think more options are better - I'll rework the inputs so that it can be either GFF or GBK

mberacochea:
Nice!. I just noticed we are not running nf-tests on Github Actions (this is not a problem with this PR BTW)

mberacochea (changes requested):
Good stuff!. I love the trimmed antismash db, the test won't fully use that because the container already has it. We need to re-build that container .. at some point.

modules/ebi-metagenomics/antismash/tests/main.nf.test
```diff
@@ -0,0 +1,119 @@
...
+                    "7.0.1"
+                    ]
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(path(process.out.gbk.get(0).get(1)).readLines()[0].contains("31398 bp")).match("gbk_input_gbk_out") },
```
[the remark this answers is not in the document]
chrisAta: Yes! It's getting it from the output, so you can check content in the first line for example like what I did there... it's a bit horrible syntax wise so hopefully it gets easier with time but it does work

modules/ebi-metagenomics/antismash/main.nf
```diff
@@ -0,0 +1,82 @@
...
+
+        tar -czf ${prefix}_antismash.tar.gz ${prefix}_results
+
+        cat <<-END_VERSIONS > versions.yml
+        "${task.process}":
+            antiSMASH: \$(echo \$(antismash --version | sed 's/^antiSMASH //' ))
+            antiSMASH database: ${db_version}
+        END_VERSIONS
+        """
+    }
```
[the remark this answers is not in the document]
chrisAta: very nice, this worked! replaced it

mberacochea (approved):
:shipit: thanks @chrisAta

PR #45, opened 2024-06-13
New module: hmmer/hmmsearch

chrisAta:


modules/ebi-metagenomics/hmmer/hmmsearch/main.nf
```diff
@@ -0,0 +1,67 @@
+
+process HMMER_HMMSEARCH {
+    tag "$meta.id"
+    label 'process_single'
```
vagkaratzas, line 4: I would say process_medium, since most often we'll give a bunch of CPUs and it takes a while for big datasets

modules/ebi-metagenomics/hmmer/hmmsearch/main.nf
```diff
@@ -0,0 +1,67 @@
...
+                break
+            }
+        }
+    }
+
+    """
+    hmmsearch \\
+        ${args} \\
+        --cpu ${task.cpus} \\
+        --domtblout ${prefix}_hmmsearch.tbl \\
```
vagkaratzas, line 44: I would go for ${prefix}_hmmsearch.domtbl,
to differentiate from --tblout result which is usually .tbl

modules/ebi-metagenomics/hmmer/hmmsearch/meta.yml
```diff
@@ -0,0 +1,49 @@
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/meta-schema.json
+name: "hmmer_hmmsearch"
+description: search for sequences against a HMM profile database
```
vagkaratzas, line 4: I think nf-core's description is a bit more accurate here, which I would change a bit in:
"Use an HMM profile to search for hits in a sequence database"

modules/ebi-metagenomics/hmmer/hmmsearch/meta.yml
```diff
@@ -0,0 +1,49 @@
...
+output:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. `[ id:'sample1', single_end:false ]`
+  - tbl:
+      type: file
+      description: Table output file containing results of hmmsearch
+      pattern: "*.tbl"
```
vagkaratzas, line 41: .domtbl better

modules/ebi-metagenomics/hmmer/hmmsearch/meta.yml
```diff
@@ -0,0 +1,49 @@
...
+      description: |
+        Groovy Map containing sample information
+        e.g. `[ id:'sample1', single_end:false ]`
+  - fasta:
+      type: file
+      description: FASTA file containing amino acid sequences to be searched for in profile
+      pattern: "*.fasta"
+  - hmm_db:
+      type: file
+      description: HMM profile database
```
vagkaratzas, line 30: I would refrain from using the word database. Correct me if I'm wrong, but hmmsearch takes a single HMM profile every time to fish hits from a sequence db. Just "HMM profile" should work the best here I think.

modules/ebi-metagenomics/hmmer/hmmsearch/meta.yml
```diff
@@ -0,0 +1,49 @@
...
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. `[ id:'sample1', single_end:false ]`
+  - fasta:
+      type: file
+      description: FASTA file containing amino acid sequences to be searched for in profile
+      pattern: "*.fasta"
+  - hmm_db:
```
vagkaratzas, line 28: just 'hmm', see comment below

modules/ebi-metagenomics/hmmer/hmmsearch/meta.yml
```diff
@@ -0,0 +1,49 @@
...
+      licence: ["BSD"]
+input:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. `[ id:'sample1', single_end:false ]`
+  - fasta:
+      type: file
+      description: FASTA file containing amino acid sequences to be searched for in profile
```
vagkaratzas, line 26: "Pool of sequences from where hits will be search for with the HMM profile"
or something like that

modules/ebi-metagenomics/hmmer/hmmsearch/main.nf
```diff
@@ -0,0 +1,67 @@
...
+    label 'process_single'
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/hmmer:3.4--hdbdd923_1' :
+        'biocontainers/hmmer:3.4--hdbdd923_1' }"
+
+    input:
+    tuple val(meta), path(fasta)
+    path hmm_db
```
vagkaratzas, line 13: I would put this as the first input:
tuple val(meta), path(hmm)
path fasta

see comment on meta.yml

modules/ebi-metagenomics/hmmer/hmmsearch/main.nf
```diff
@@ -0,0 +1,67 @@
...
+
+    output:
+    tuple val(meta), path("*.tbl"), emit: tbl
```
vagkaratzas, line 16: do:
path("*.domtbl"), emit: domtbl,
to differentiate from results of --tblout

modules/ebi-metagenomics/hmmer/hmmsearch/main.nf
```diff
@@ -0,0 +1,67 @@
...
+        hmmer: \$(hmmsearch -h | grep 'HMMER' | cut -d' ' -f3)
+    END_VERSIONS
+    """
+
+    stub:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+
+    """
+    touch ${prefix}_hmmsearch.tbl
```
vagkaratzas, line 60: .domtbl

modules/ebi-metagenomics/hmmer/hmmsearch/main.nf
```diff
@@ -0,0 +1,67 @@
...
+        hmm_file = hmm_db
+    }
+    else{
+        for(iter_file in hmm_db) {
+            if (iter_file.getExtension() == 'hmm'){
+                hmm_file = iter_file
+                break
+            }
+        }
+    }
```
vagkaratzas, line 38: I don't understand the purpose of the block of code above; if multiple HMMs are passed here in a list, we only keep the first one and then run hmmsearch on that? I am not sure it will work properly for a list in this way, and I don't see an nf-test with this scenario.
This could probably be done with Groovy code outside this module (or another pre-filtering module before this one) and then pass one HMM profile as input in HMMER_HMMSEARCH

vagkaratzas (changes requested):
In general, the existing module in nf-core should work fine, but I also hate the fact that they zip the results without an option to disable that. So I am up for making our own module for this.
What I don't get is the fact that we are trying to pass a list of HMMs in an HMM db -but then only keep one HMM? hmmsearch should search with one profile against a fasta file from what I know. See comments and let's try to resolve this. Unless I completely missed the mark here.

chrisAta:
Thanks for the amazing feedback @vagkaratzas ! I'm going to close this PR now, since I thought that hmmsearch could use the index files that come from hmmpress,  but actually those are only for hmmscan not hmmsearch. That was the main reason I was thinking of making a new module instead of using the nf-core one, so I can just fall back to that now 😄

PR #47, opened 2024-06-27
New Module: genome-properties

chrisAta:
Part of the series of modules we need for V6 Assembly Analysis


modules/ebi-metagenomics/genomeproperties/main.nf
```diff
@@ -0,0 +1,59 @@
+
+process GENOMEPROPERTIES {
+    tag "$meta.id"
+    label 'process_single'
+
+    conda "${moduleDir}/environment.yml"
+    container 'microbiome-informatics/genome-properties:2.0'
```
mberacochea, line 7: Can we `bioconda` this one?

modules/ebi-metagenomics/genomeproperties/main.nf
```diff
@@ -0,0 +1,59 @@
...
+        -gpff genomeProperties.txt \\
+        -name ${prefix} \\
+
+    mv JSON_${prefix} ${prefix}_gp.json
+    mv SUMMARY_FILE_${prefix} ${prefix}_gp.txt
+    mv TABLE_${prefix} ${prefix}_gp.tsv
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        genomeproperties: ${gp_version}
```
mberacochea, line 40:
```suggestion
        Genome Properties: ${gp_version}
```

modules/ebi-metagenomics/genomeproperties/tests/nextflow.config
```diff
@@ -0,0 +1,7 @@
+process{
+
+    withName: GENOMEPROPERTIES {
+        ext.args = '-all -outfiles table -outfiles web_json -outfiles summary'
```
mberacochea, line 4: Why not having these in the main script?

mberacochea (changes requested):
Nice, thanks @chrisAta I left some comments.
We "own" the repo so we should try and get this one into bioconda too.

modules/ebi-metagenomics/genomeproperties/main.nf
chrisAta: sure we could! I'll add it to my tasks

modules/ebi-metagenomics/genomeproperties/tests/nextflow.config
chrisAta: Because these are optional - in V5 we do output all of `-outfiles` parameters for example, but it could be we decide to only output one in the future so leaving that flexibility

chrisAta:
> Nice. Some minor typos only. But I was also wondering; would it make sens (i.e. produce results) if instead of 1000 lines test file you fed it a 10 line one? Less storage space, faster test.

Thank you! I'd rather keep it like this, the test output is more realistic and the test finishes almost instantly, also the input file is very small (223kb), but good to think about for the future

modules/ebi-metagenomics/genomeproperties/main.nf
mberacochea: Thanks!!

mberacochea (approved):
:shipit:

PR #48, opened 2024-06-27
First commit for DRAM distill

Ge94:
I started from [Ales' implementation](https://github.com/EBI-Metagenomics/shallowmapping/blob/dev/modules/local/dram/distill.nf) and tried to convert this to an nf-core fashion. If you could give me any advice on how to correct this (or confirm whether I am on the right track) I would be really grateful.

I have a couple things that I'd like you to address if possible! 

1. I followed [nf-core guidelines](https://nf-co.re/docs/guidelines/components/modules) as much as I could. The only thing left for sure is [this part](https://nf-co.re/docs/guidelines/components/modules#use-of-conda-directive) - is it needed?
2. I wasn't sure about how to address dbs, especially when testing. Right now they are set as [fixtures](https://github.com/EBI-Metagenomics/nf-modules/tree/dram_distill/modules/ebi-metagenomics/dram/distill/tests/fixtures/dram_distill_dbs) since they are not very big (1.7 MB). 
3. Ale calls her module 4 times in her pipeline, and these 4 [output files are named based on different parameters](https://github.com/EBI-Metagenomics/shallowmapping/blob/dev/modules/local/dram/distill.nf#L44). I got rid of this as it's my understanding that everything should be passed via args/prefix, should we update her pipeline accordingly?
- If so, probably we should rearrange how she generates files (currently some of it is in the distill module)

Thank you lots!


modules/ebi-metagenomics/dram/distill/main.nf
```diff
@@ -0,0 +1,54 @@
+process DRAM_DISTILL {
+    tag "$meta.id"
+    label 'process_high'
+
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/dram:1.3.5--pyhdfd78af_0':
+        'quay.io/biocontainers/dram:1.3.5--pyhdfd78af_0' }"
+
+    containerOptions="--bind ${moduleDir}/tests/fixtures/dram_distill_dbs/:/data/ --bind ${moduleDir}/tests/fixtures/dram_distill_dbs/CONFIG:/usr/local/lib/python3.10/site-packages/mag_annotator/CONFIG"
```
chrisAta, line 9: I assume you absolutely need this data mounted at those paths for DRAM to work? if so this makes sense, though slight change I would do. At the moment this would only work with singularity (`--bind`) but not with docker (`-v`). You can do something like this:

```
    containerOptions {
        if (workflow.containerEngine == 'singularity') {
            return "--bind ${interproscan_db}/data:/opt/interproscan-5.62-94.0/data"
        } else {
            return "-v ${interproscan_db}/data:/opt/interproscan-5.62-94.0/data"
        }
    }
```

Just replace the interproscan stuff with dram :)

modules/ebi-metagenomics/dram/distill/main.nf
```diff
@@ -0,0 +1,54 @@
...
+
+    input:
+    tuple val(meta), path(dram_input)
```
chrisAta, line 12: I would maybe rename dram_input to something more meaningful depending on its format? is it a tsv, fasta, etc? one example could be `tsv_input`

modules/ebi-metagenomics/dram/distill/main.nf
```diff
@@ -0,0 +1,54 @@
...
+
+    output:
+    tuple val(meta), path("*_dram.html"), emit: distill_out_html // removed "optional" e.g. we don't launch the module when missing input files
+    tuple val(meta), path("*_dram.tsv") , emit: distill_out_tsv
```
chrisAta, line 16: I think here you don't need the `distill` prefix for each output, as it's a self-contained module that will only output dram output anyway. you can rename these to `html` and `tsv` IMO

modules/ebi-metagenomics/dram/distill/main.nf
```diff
@@ -0,0 +1,54 @@
...
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def VERSION = '1.3.5' // WARN: dram has no option to print the tool version. This is the container version
+
+    """
+    DRAM.py \\
+        distill \\
+        -i ${dram_input}  \\
+        -o dram_out
```
chrisAta, line 31: not sure if DRAM has any of these, but if it can have other parameters that are optional, this is where you should add $args as an extra line, makes it easy for users to add any extra arguments if they wish

modules/ebi-metagenomics/dram/distill/meta.yml
```diff
@@ -0,0 +1,42 @@
...
+      doi: "doi.org/10.1093/nar/gkaa621"
+      licence: ["GPL v3"]
+
+  input:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. [ id:'test', single_end:false ]
+  - dram_input:
```
chrisAta, line 22: if you do change names of inputs/outputs, don't forget to change them here too

modules/ebi-metagenomics/dram/distill/meta.yml
```diff
@@ -0,0 +1,42 @@
...
+      type: file
+      description: Input tsv file containing sequence annotations from different tools
+      pattern: "*.{tsv}"
+
+output:
+  - distill_out_html:
+      type: file
+      description: HTML file containing heatmaps of summarised annotations
+      pattern: "*.{html}"
```
chrisAta, line 31: same as last comment

chrisAta (changes requested):
This looks really good and almost there! Just a few changes needed. The biggest one is more of a location switch - I would move all of the unit test related things into the actual module directory. That will likely mean you can just delete the custom `nf-test.config` file you wrote to connect the two directories

mberacochea (changes requested):

mberacochea:
> I started from [Ales' implementation](https://github.com/EBI-Metagenomics/shallowmapping/blob/dev/modules/local/dram/distill.nf) and tried to convert this to an nf-core fashion. If you could give me any advice on how to correct this (or confirm whether I am on the right track) I would be really grateful.
> 
> I have a couple things that I'd like you to address if possible!
> 
>     1. I followed [nf-core guidelines](https://nf-co.re/docs/guidelines/components/modules) as much as I could. The only thing left for sure is [this part](https://nf-co.re/docs/guidelines/components/modules#use-of-conda-directive) - is it needed?
> 
>     2. I wasn't sure about how to address dbs, especially when testing. Right now they are set as [fixtures](https://github.com/EBI-Metagenomics/nf-modules/tree/dram_distill/modules/ebi-metagenomics/dram/distill/tests/fixtures/dram_distill_dbs) since they are not very big (1.7 MB).
> 
>     3. Ale calls her module 4 times in her pipeline, and these 4 [output files are named based on different parameters](https://github.com/EBI-Metagenomics/shallowmapping/blob/dev/modules/local/dram/distill.nf#L44). I got rid of this as it's my understanding that everything should be passed via args/prefix, should we update her pipeline accordingly?
> 
> 
>     * If so, probably we should rearrange how she generates files (currently some of it is in the distill module)
> 
> 
> Thank you lots!

It's looke pretty good to me :). I left one comment about the database mounts.

Answers to your specific questions:
1) We don't use conda, so we have been ignoring that one. I think we should start adding it and pushing the modules to nf-core. Not now, or but in the short term future.
2) 1.7MB it's alright. Do the tests with this db produce meaningful results, or just empty results?.
3) Yes, that is just a specific thing of that pipeline

Ge94:
I believe that after a brief discussion on slack we are alright with the points above?

WRT the test output question, the test produces an actual result. I just made it a bit prettier.

modules/ebi-metagenomics/dram/distill/main.nf
```diff
@@ -0,0 +1,61 @@
...
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def VERSION = '1.3.5' // WARN: dram has no option to print the tool version. This is the container version
+
+    """
+    DRAM.py \\
+        distill \\
+        -i ${tsv_input}  \\
+        -o dram_out \\
+        ${args}
+    mv dram_out/product.html ${prefix}_dram.html
```
mberacochea, line 39: DRAM produces several .html files if the number of annotations / genome size is beyond a certain threshold.. I fixed shallow mapping pipeline -> https://github.com/EBI-Metagenomics/shallowmapping/commit/4f13041d74be04711f8d4d2659c93e44bfc5b3c2

Ge94: Looking at dram distill's code - looks like the default threshold is 1000 and it can be adjusted with a flag
https://github.com/WrightonLabCSU/DRAM/blob/03cb821ed9c37793ca886eb14c7740fd24712af7/scripts/DRAM.py#L150

Ales-ibt: I haven't seen an example of DRAM producing more than one HTML output per sample. However, in real life we will have to process big assemblies that can probably yield a large number of annotations. Worthy to test maybe with one of the chicken-gut huge assemblies just to check behaviour.

modules/ebi-metagenomics/dram/distill/tests/fixtures/dram_dbs/bkp_CONFIG
Ales-ibt, line 1: Do we need this file?

Ales-ibt (approved):
Thank you for this. I think just keep an eye on the multiple html outputs that could screw up the renaming step when the outputs are published.

modules/ebi-metagenomics/dram/distill/main.nf
mberacochea: I think Jen bumped into this, the assembly that I'm using for testing [ASA](https://github.com/EBI-Metagenomics/assembly-analysis-pipeline) it's pretty big so it would be a nice test case.

modules/ebi-metagenomics/dram/distill/main.nf
```diff
@@ -0,0 +1,73 @@
...
+        : 'quay.io/biocontainers/dram:1.3.5--pyhdfd78af_0'}"
+
+    containerOptions {
+        def arg = "--volume"
+        if (workflow.containerEngine == 'singularity' || workflow.containerEngine == 'apptainer') {
+            arg = "--bind"
+        }
+        def mounts = [
+            "${params.reference_dbs}/dram_dbs/:/data/",
+            "${params.reference_dbs}/dram_dbs/DRAM_CONFIG.json:/usr/local/lib/python3.10/site-packages/mag_annotator/CONFIG",
```
mberacochea, line 16: I think it's better to have the reference_dbs as a an input.. because otherwise you are forcing anyone that want to use this module to use that parameter name (which is very generic) or to modify the module

modules/ebi-metagenomics/dram/distill/main.nf
```diff
@@ -0,0 +1,73 @@
...
+    DRAM.py \\
+        distill \\
+        -i ${tsv_input}  \\
+        -o dram_out \\
+        ${args}
+
+    mv dram_out/product.html ${prefix}_dram.html
+    mv dram_out/product.tsv ${prefix}_dram.tsv
+    mv dram_out/genome_stats.tsv ${prefix}_genome_stats.tsv
+    mv dram_out/metabolism_summary.xlsx ${prefix}_metabolism_summary.xlsx
```
mberacochea, line 50: We are gzipping the results in most modules

modules/ebi-metagenomics/dram/distill/main.nf
```diff
@@ -0,0 +1,73 @@
...
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    // WARN: dram has no option to print the tool version. This is the container version
+    def VERSION = '1.3.5' 
+
+    """
+    DRAM.py \\
+        distill \\
+        -i ${tsv_input}  \\
```
mberacochea, line 43: Please handle a .gz input too

mberacochea (changes requested):
Thank you @Ge94 !

mberacochea (approved):
:shipit:

PR #49, opened 2024-07-03
Refactor/eggnogmapper

chrisAta:
Refactor eggnogmapper to have less hardcoded things. This should allow it to be used to for annotation and mapping depending on the supplied `$args` like here: https://github.com/EBI-Metagenomics/mettannotator/blob/main/modules/local/eggnog.nf


modules/ebi-metagenomics/eggnogmapper/meta.yml
```diff
@@ -27,6 +27,11 @@ input:
       description: Database of sequences in FASTA format
       pattern: "*.{fasta,fa,fasta.gz,fa.gz}"
```
vagkaratzas, line 28: If you are removing the support for compressed files, you'll have to remove it from here as well

modules/ebi-metagenomics/eggnogmapper/main.nf
```diff
@@ -9,38 +9,40 @@ process EGGNOGMAPPER {
 
     input:
     tuple val(meta), path(fasta)
-    path(eggnog_db)
+    tuple val(meta), path(annotation_hit_table)
```
vagkaratzas, line 12: Better rename this to meta2. Won't be always the same

modules/ebi-metagenomics/eggnogmapper/main.nf
```diff
@@ -9,38 +9,40 @@ process EGGNOGMAPPER {
...
+    path "versions.yml"                    , emit: versions
 
     when:
     task.ext.when == null || task.ext.when
 
     script:
     def args = task.ext.args ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
-    def is_compressed = fasta.name.endsWith(".gz")
-    def fasta_name = fasta.name.replace(".gz", "")
```
vagkaratzas, line 27: Is it difficult to maintain the compressed fasta support with the new functionality?

modules/ebi-metagenomics/eggnogmapper/meta.yml
```diff
@@ -48,16 +53,23 @@ output:
...
         e.g. `[ id:'test', single_end:false ]`
-
-  - csv:
+  - hits:
       type: file
-      description: CSV containing protein annotations
-      pattern: "*.emapper.hits"
-
+      description: Tabulated file containing hits
+      pattern: "*.hits"
```
vagkaratzas, line 59: I think *emapper.hits here (as seen in main.nf)...

modules/ebi-metagenomics/eggnogmapper/main.nf
```diff
@@ -9,38 +9,40 @@ process EGGNOGMAPPER {
...
     path(eggnog_data_dir)
+    path(eggnog_db)
     path(eggnog_diamond_db)
 
     output:
-    tuple val(meta), path("*.emapper.hits"), emit: csv
-    path "versions.yml",                     emit: versions
+    tuple val(meta), path("*.emapper.hits"), emit: hits, optional: true
+    tuple val(meta), path("*annotations*") , emit: annotations, optional: true
+    tuple val(meta), path("*orthologs*")   , emit: orthologs, optional: true
```
vagkaratzas, line 20: Make sure we need the * after annotations and orthologs (and not in the .emapper.hits). I don't remember the actual produced file extensions

modules/ebi-metagenomics/eggnogmapper/meta.yml
```diff
@@ -48,16 +53,23 @@ output:
...
+  - annotations:
+      type: file
+      description: Tabulated file containing annotations
+      pattern: "*.annotations"
+  - orthologs:
+      type: file
+      description: Tabulated file containing EggNOG orthologs
+      pattern: "*.emapper.orthologs"
```
vagkaratzas, line 67: ...and *orthologs here? Just make sure you follow the same format as in the main.nf outputs

modules/ebi-metagenomics/eggnogmapper/tests/main.nf.test
```diff
@@ -0,0 +1,99 @@
...
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
+            )
+        }
+
+    }
+
```
vagkaratzas, line 98: Can I get you into the trouble of also making the stub nf-test? These are also mandatory for nf-core modules now. Just following the guidelines :D

modules/ebi-metagenomics/eggnogmapper/tests/main.nf.test
```diff
@@ -0,0 +1,99 @@
...
+                process {
+                    """
+                    input[0] = [
+                        [ id:'test', single_end:false ], // meta map
+                        file("${moduleDir}/tests/fixtures/test.fasta", checkIfExists: true)
+                        ]
+                    """
+                }
+            }
+        }
```
vagkaratzas, line 69: remove setup from this test, diamond not needed

vagkaratzas (changes requested):
Awesome work with the inputs, thanks!! I am not sure how the annotation input mode works, but I see it only uses the eggonog data dir, and not the eggong db itself. And probably it's either the fasta file, for the first mode, or the annotation file, for the second mode. Left some comments here and there!

modules/ebi-metagenomics/eggnogmapper/main.nf
```diff
@@ -9,38 +9,48 @@ process EGGNOGMAPPER {
...
-    path(eggnog_db)
+    tuple val(meta2), path(annotation_hit_table)
     path(eggnog_data_dir)
+    path(eggnog_db)
     path(eggnog_diamond_db)
 
     output:
-    tuple val(meta), path("*.emapper.hits"), emit: csv
-    path "versions.yml",                     emit: versions
+    tuple val(meta), path("*.hits")   , emit: hits, optional: true
```
vagkaratzas, line 18: don't forget the extra empty spaces behind the comma now :D

vagkaratzas (approved):
Excellent!

modules/ebi-metagenomics/eggnogmapper/main.nf
chrisAta: hahahahahaha there will always be at least _one_ of these I forget it seems 😄 thanks for the review!

PR #50, opened 2024-07-05
New module: kegg-pathways-completeness

chrisAta:
Use module to use the kegg-pathways-completeness tool, recently published to bioconda


modules/ebi-metagenomics/keggpathwayscompleteness/main.nf
```diff
@@ -0,0 +1,77 @@
...
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/kegg-pathways-completeness:1.0.4--pyhdfd78af_0':
+        'biocontainers/kegg-pathways-completeness:1.0.4--pyhdfd78af_0' }"
+
+    input:
+    tuple val(meta), path(fasta)
+    tuple val(meta), path(hmmsearch_tbl)
+    tuple val(meta), path(ko_list)
```
mberacochea, line 14: I think if the 3 inputs need the meta, maybe there should only be one input: tuple val(meta), path(fasta), path(hmmsearch_tbl), path(ko_list) as otherwise using this to process "many" samples won't work as it won't be possible to process the same sample with 3 different inputs.

modules/ebi-metagenomics/keggpathwayscompleteness/main.nf
```diff
@@ -0,0 +1,77 @@
...
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def version = "1.0.4" // No way to get version automatically so hard-coding it
+
+    if (ko_list == []){
+        """
+        sed \\
+        '/^#/d; s/ \\+/\\t/g' \\
```
mberacochea, line 32: Would you mind adding a comment about this filter? Why do we need it?

modules/ebi-metagenomics/keggpathwayscompleteness/main.nf
```diff
@@ -0,0 +1,77 @@
...
+    }
+    else {
+        """
+        give_pathways \\
+        -l ${ko_list} \\
+        -o ${prefix}
+
+        cat <<-END_VERSIONS > versions.yml
+        "${task.process}":
+            keggpathwayscompleteness: ${version}
```
mberacochea, line 58:
```suggestion
            kegg-pathways-completeness: ${version}
```
I would use the bioconda and pypi package name.

modules/ebi-metagenomics/keggpathwayscompleteness/meta.yml
```diff
@@ -0,0 +1,56 @@
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/meta-schema.json
+name: "keggpathwayscompleteness"
```
mberacochea, line 3:
```suggestion
name: "kegg-pathways-completeness"
```

mberacochea (changes requested):

KateSakharova (changes requested):
As we discussed with Chris I think we should only use `give_pathways` command in that module (as it is required in tool's README). Then we can re-use it in v6 and shallow-shotgun pipeline

modules/ebi-metagenomics/keggpathwayscompleteness/main.nf
```diff
@@ -0,0 +1,51 @@
...
+    tuple val(meta), path("*_pathways.tsv"), emit: kegg_pathways
+    path "versions.yml"                    , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def kegg_input = ko_list ? "-l ${ko_list}" : "-i ${filtered_tbl}"
```
KateSakharova, line 25: I'm wondering is it possible to add a WARN or check if both files are presented as input..

KateSakharova (approved):
I like it! Thank you, Chris!

modules/ebi-metagenomics/keggpathwayscompleteness/main.nf
chrisAta: I added a check:
```
if (ko_list && filtered_tbl){
        log.warn("Both \$ko_list and \$filtered_tbl were given as input types, will fall back to using \$filtered_tbl i.e. ${filtered_tbl}");
    }
```

Which prints this in the nextflow.log if it's true:
`Jul-09 10:20:56.720 [Actor Thread 2] WARN  nextflow.Nextflow - Both $ko_list and $filtered_tbl were given as input types, will fall back to using $filtered_tbl i.e. test_filtered.tbl`

KateSakharova: but you will use `ko_list` if both are presented, don't you?

modules/ebi-metagenomics/keggpathwayscompleteness/main.nf
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    """
+    give_pathways \\
+    ${kegg_input} \\
+    -o ${prefix}
+
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        kegg-pathways-completeness: \$(give_pathways --version | cut -d' ' -f2)
```
mberacochea, line 38: image

mberacochea (approved):
When you address Kate's comment, I'm happy for this one to be merged

PR #51, opened 2024-08-06
New Module: fastqutils

chrisAta:
New fastq_utils module, that uses the fastq_info command to perform initial sanity-checking of FASTQ files. Have written unit tests with dummy data that checks the response of every kind of error the module is able to identify


mberacochea (approved):
I think we are going to opt for SeqFu, but this module it's ready... it may come in handy in the future.. so :shipit:

PR #53, opened 2024-08-23
New module: library_strategy_check

chrisAta:


modules/ebi-metagenomics/librarystrategycheck/meta.yml
```diff
@@ -0,0 +1,47 @@
...
+
+output:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. `[ id:'sample1', single_end:false ]`
+  - txt:
+      type: file
+      description: Text file containing the string "AMPLICON" if classified as AMPLICON, empty file otherwise
```
mberacochea, line 37: We should use env(is_amplicon) instead of a file, maybe we can store the file anyway (to make it easier to debug). Like https://github.com/EBI-Metagenomics/mettannotator/blob/main/modules/local/lookup_kingdom.nf#L13

mberacochea (changes requested):
My main feedback would be to combine `assess_mcp_proportions` and this module into a single PROCESS.

modules/ebi-metagenomics/librarystrategycheck/meta.yml
```diff
@@ -0,0 +1,47 @@
...
+
+input:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. `[ id:'sample1', single_end:false ]`
+  - mcp_cons:
+      type: file
+      description: MCP conservation curve file output by `assess_mcp_proportions` tool, also from the mgnify-pipelines-toolkit
```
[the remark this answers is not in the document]
chrisAta: Agreed, I'll bring in the existing `assess_mcp_proportions` branch into here and make them into one module

modules/ebi-metagenomics/assessmcpproportions/main.nf
```diff
@@ -0,0 +1,97 @@
...
+        strands = "R"
+    }
+
+    if (strands == "") {
+        """
+        touch ${prefix}_${var_region}_mcp_cons.tsv
+
+        cat <<-END_VERSIONS > versions.yml
+        "${task.process}":
+            mgnify-pipelines-toolkit: 0.1.5
```
mberacochea, line 45: Can't we use "--version"?

modules/ebi-metagenomics/assessmcpproportions/main.nf
```diff
@@ -0,0 +1,97 @@
...
+            -o ./
+
+        check_out=\$(cat ${prefix}_library_check_out.txt)
+
+        cat <<-END_VERSIONS > versions.yml
+        "${task.process}":
+            mgnify-pipelines-toolkit: 0.1.5
+        END_VERSIONS
+        """
+    }   else {
```
mberacochea, line 68:
```suggestion
    } else {
```

mberacochea (changes requested):
2 very small comments... but other than that it looks pretty good to me :)

mberacochea (approved):
Brilliant! Thank you @chrisAta

PR #61, opened 2024-09-06
Module/fastq suffix header check

chrisAta:


modules/ebi-metagenomics/fastqsuffixheadercheck/main.nf
```diff
@@ -0,0 +1,67 @@
...
+        -r ${fastq[1]} \\
+        -s ${prefix} \\
+        -o ./
+
+        cat <<-END_VERSIONS > versions.yml
+        "${task.process}":
+            mgnify-pipelines-toolkit: \$(get_mpt_version)
+        END_VERSIONS
+        """
+    }
```
vagkaratzas, line 51: For cleaner code without duplication, can you maybe put either -f ${fastq} (for signle read) or -f ${fastq[0]} -r ${fastq[1]} (if paired) in a Groovy variable, and then run only one script instead? This will result in one command + one versions output

modules/ebi-metagenomics/fastqsuffixheadercheck/main.nf
```diff
@@ -0,0 +1,67 @@
...
+        END_VERSIONS
+        """
+
+    } else{
+        """
+        fastq_suffix_header_check \\
+        -f ${fastq[0]} \\
+        -r ${fastq[1]} \\
+        -s ${prefix} \\
+        -o ./
```
vagkaratzas, line 44: push these four lines above one tab inside ;)

modules/ebi-metagenomics/fastqsuffixheadercheck/main.nf
chrisAta: nice idea! did this:
`def fastq_input = meta.single_end ? "-f ${fastq}" : "-f ${fastq[0]} -r ${fastq[1]}"`

vagkaratzas (approved):

PR #63, opened 2024-10-08
Add sanity checking steps to reads_qc subworkflow

chrisAta:
This PR adds sanity checking using three new modules before running fastp


KateSakharova (changes requested):
Nice work! 
That subwf is amplicon focused. I think we should make it more universal. Maybe it is required a flag like val(filter_amplicon) and wrap amplicon check into if-else. We can also add more similarar filters for WGS/RNA-Seq if we need.

KateSakharova (approved):

subworkflows/ebi-metagenomics/reads_qc/main.nf
```diff
@@ -1,35 +1,108 @@
...
+        }
+    }
+    // ***** Necessary mapping functions *****
+
+    SEQFU_CHECK(ch_reads)
+    ch_versions = ch_versions.mix(SEQFU_CHECK.out.versions.first())
+
+    passed_seqfu_reads = SEQFU_CHECK.out.tsv
+        .splitCsv(sep: "\t", elem: 1)
+        .map(filterBySeqFuStatus)
```
mberacochea, line 56: I don't think map can be used to filter things.. for that [`.filter` is the right operator](https://www.nextflow.io/docs/latest/reference/operator.html#filter)

I suspect this map is leaving empty values that then are filtered in the .joins, but it's a bit dirty IMO.

mberacochea: Like -> https://github.com/EBI-Metagenomics/nf-modules/pull/68

mberacochea (changes requested):

subworkflows/ebi-metagenomics/reads_qc/main.nf
chrisAta: Done!

PR #65, opened 2024-10-16
CAT_pack contigs taxonomy assignation module

mberacochea:


.github/workflows/code-linting.yml
```diff
@@ -1,4 +1,5 @@
-name: Code Linting
+name: pre-commit
+
```
vagkaratzas, line 2: Should changes in this file be in this PR?

modules/ebi-metagenomics/cat/environment.yml
```diff
@@ -0,0 +1,8 @@
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/environment-schema.json
+name: "cat"
```
vagkaratzas, line 3: name is not used anymore, linting with fail with it

modules/ebi-metagenomics/cat/main.nf
```diff
@@ -0,0 +1,71 @@
...
+
+    input:
+    tuple val(meta), path(contigs_fasta)
+    path cat_db
+    path taxonomy_db
+
+    output:
+    tuple val(meta), path('*_summary.txt'), emit: summary
+    tuple val(meta), path('*_contig2classification_official_names.tsv'), emit: contigs_classification
+    path "versions.yml", emit: versions
```
vagkaratzas, line 18: harshil alignment of commas please :D

modules/ebi-metagenomics/cat/meta.yml
```diff
@@ -0,0 +1,65 @@
...
+  - contigs_fasta:
+      type: file
+      description: The fasta file that contains contigs
+      pattern: "*.{fa,fasta}"
+
+  - cat_db:
+      type: directory
+      description: Path to directory that contains CAT database files
+
+  - tax_db:
```
vagkaratzas, line 34: rename to taxonomy_db (as in channel above, or the other way around)

vagkaratzas (changes requested):
Thanks! Minor comments and let's see why tests fail

.github/workflows/code-linting.yml
mberacochea: yeah

modules/ebi-metagenomics/cat/environment.yml
mberacochea: it doesn't fail, the schema says that the name is not required

modules/ebi-metagenomics/cat/meta.yml
mberacochea: done

KateSakharova (changes requested):
Nice work!
The only thing is a name of module.. I would not call it CAT, maybe CAT_taxonomy or smth, because CAT should be more general module but we have contigs,add_names and summarise

mberacochea:
> Nice work! The only thing is a name of module.. I would not call it CAT, maybe CAT_taxonomy or smth, because CAT should be more general module but we have contigs,add_names and summarise

That is a very good point, OK. I'll rename the module

mberacochea:
> Nice work! The only thing is a name of module.. I would not call it CAT, maybe CAT_taxonomy or smth, because CAT should be more general module but we have contigs,add_names and summarise

done, renamed to CATCONTIGS.. which alings with CAT modes (contigs, bins and reads)

KateSakharova (approved):
Looks good to me!

modules/ebi-metagenomics/assessmcpproportions/main.nf
chrisAta, line 1: nice catch, no need to change anything but just wanted to say - I want to find a better way of updating the version for all `mpt` modules whenever there's a new version... soon:tm:

modules/ebi-metagenomics/assessmcpproportions/main.nf
```diff
@@ -89,6 +89,8 @@ process ASSESSMCPPROPORTIONS {
     """
     touch ${assess_mcp_prop_prefix}_mcp_cons.tsv
 
+    echo 'dummy' > ${assess_mcp_prop_prefix}_mcp_cons.tsv
```
chrisAta, line 92: that's fine but is there a reason you prefer to keep it non-empty?

modules/ebi-metagenomics/catcontigs/meta.yml
```diff
@@ -0,0 +1,67 @@
...
+    - contigs_fasta:
+        type: file
+        description: The fasta file that contains contigs
+        pattern: "*.{fa,fasta}"
+  - - cat_db:
+        type: directory
+        description: Path to directory that contains CAT database files
+  - - taxonomy_db:
+        type: directory
+        description: Path to directory that contains CAT taxonomy files
```
chrisAta, line 34: There are double dashes `- -` on these

modules/ebi-metagenomics/combinedgenecaller/merge/meta.yml
```diff
@@ -1,82 +1,82 @@
...
-      type: file
-      description: protein translations file
-      pattern: "*.{faa}"
-
-  - mask_file:
-      type: file
-      description: sequence mask file (cmsearch tblout)
-      pattern: "*.*"
-
+  - - meta:
```
chrisAta, line 19: double dash

chrisAta (approved):
Looks good to me! Some questions and tiny changes

modules/ebi-metagenomics/assessmcpproportions/main.nf
mberacochea: nf-core liniting don't like empty files (which overall I think it's good)
Having stubs with some content is useful IMO

modules/ebi-metagenomics/catcontigs/meta.yml
mberacochea: To represent tuples, like
```
input:
tuple val(meta), path(thingy)
path(db)
```
used to be represented as a flat schema, now it should be represented as

```
input:
 - #inputone
  - meta
  - thingy
 - db
```

modules/ebi-metagenomics/combinedgenecaller/merge/meta.yml
mberacochea: yup, explain in the previuos comment

modules/ebi-metagenomics/assessmcpproportions/main.nf
mberacochea: BTW, this is not a good example.. a real piece of data on this file would be useful

modules/ebi-metagenomics/catcontigs/meta.yml
chrisAta: oooh fancy!! I didn't know, thank you

modules/ebi-metagenomics/assessmcpproportions/main.nf
mberacochea: A Github bot thingy? - perfect for a mini-hackathon track

modules/ebi-metagenomics/catcontigs/meta.yml
```diff
@@ -0,0 +1,67 @@
...
+          Groovy Map containing sample information
+          e.g. `[ id:'sample1', single_end:false ]`
+    - contigs_fasta:
+        type: file
+        description: The fasta file that contains contigs
+        pattern: "*.{fa,fasta}"
+  - - cat_db:
+        type: directory
+        description: Path to directory that contains CAT database files
+  - - taxonomy_db:
```
chrisAta, line 32: from your explanation, shouldn't these have only one dash then? since they are:
```
    path cat_db
    path taxonomy_db
```
as inputs, no tuple

vagkaratzas (approved):
Let's merge this and see. If CAT will contain more submodules in the future, we'll need to rename to cat/contigs

modules/ebi-metagenomics/catcontigs/meta.yml
mberacochea: well... the linter doesn't like that:
```
The meta.yml of the module catcontigs is not valid: {'cat_db': {'type': 'directory', 'description': 'Path to directory that contains CAT database files'}} is not of type 'array'. Check the entry for input.
```

:sigh: I leave this one as it is, nf-core + nextflow are working on this (trying to use the schemas directly in nextflow) so let's wait

PR #71, opened 2025-01-23
Add owltools module

chrisAta:


modules/ebi-metagenomics/owltools/meta.yml
```diff
@@ -0,0 +1,54 @@
...
+input:
+  - - meta:
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. `[ id:'sample1', single_end:false ]`
+    - input_gaf:
+        type: file
+        description: Input GAF file
+        pattern: "*.gaf"
```
mberacochea, line 26: This GAF files it's generated by one the toolkit scripts, right? I need to add step in pipeline to do so.

modules/ebi-metagenomics/owltools/meta.yml
```diff
@@ -0,0 +1,54 @@
...
+      licence: ["BSD-3-Clause"]
+input:
+  - - meta:
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. `[ id:'sample1', single_end:false ]`
+    - input_gaf:
+        type: file
+        description: Input GAF file
```
mberacochea, line 25:
```suggestion
        description: Input Gene Association File (GAF) file
```
if I'm not mistaken

mberacochea (approved):
Not much to add, one suggestion on adding a bit of a description to the file formats (specially those that are not common like GAF)

modules/ebi-metagenomics/owltools/meta.yml
mberacochea: nop, it's http://geneontology.org/page/go-annotation-file-gaf-format-21

modules/ebi-metagenomics/owltools/meta.yml
mberacochea: Ignore this.. :dumb: from me. This is reason for this module to exist.

PR #75, opened 2025-01-27
Adds the mgnify-pipelines-toolkit -> add_rhea_chebi_annotation module

mberacochea:


modules/ebi-metagenomics/mgnifypipelinestoolkit/rheachebiannotation/main.nf
```diff
@@ -0,0 +1,59 @@
+process MGNIFYPIPELINESTOOLKIT_RHEACHEBIANNOTATION {
```
KateSakharova, line 1: @chrisAta look at those name s :fear: 
All other toolkit modules are named without **MGNIFYPIPELINESTOOLKIT**

ochkalova, line 1: 🤯

chrisAta: oof! does it need the whole name to it matches the parent directory? if not if we really want to have a prefix for these modules we could do
```suggestion
process MPT_RHEACHEBIANNOTATION {
```
instead?

modules/ebi-metagenomics/mgnifypipelinestoolkit/rheachebiannotation/meta.yml
```diff
@@ -0,0 +1,67 @@
...
+output:
+  - rhea2proteins_tsv:
+      - meta:
+          type: map
+          description: |
+            Groovy Map containing sample information
+            e.g. `[ id:'sample1', single_end:false ]`
+      - "*.tsv":
+          type: file
+          description: "TSV file with columns: contig_id, protein_id, UniRef90 cluster, rhea_ids, CHEBI reaction participants"
```
ochkalova, line 53: The output columns are different...
contig_id, protein_id, protein_hash, rhea_ids, chebi_reaction, reaction_definition, is_top_hit
My bad, I haven't updated the doc strings in the script...

modules/ebi-metagenomics/mgnifypipelinestoolkit/rheachebiannotation/main.nf
```diff
@@ -0,0 +1,59 @@
...
+        gzip -c -d $fasta > ${fasta_file}
+    fi
+
+    if [ "$is_diamond_compressed" == "true" ]; then
+        gzip -c -d $diamond_tsv > ${diamond_file}
+    fi
+
+    # FIXME: fix the toolkit, this script has to be called like this otherwise it complains about the argumentss
+    python /usr/local/lib/python3.11/site-packages/mgnify_pipelines_toolkit/analysis/assembly/add_rhea_chebi_annotation.py \\
+        --input ${diamond_file} \\
```
ochkalova, line 38: Maybe make a note that `--input` will be changed to `--diamond_hits` after toolkit update?

Ge94 (approved):
Apart from the naming convention everything looks good! IMO, let's open a JIRA ticket for that and keep the name "fuzzy" for now.

ochkalova (approved):

modules/ebi-metagenomics/mgnifypipelinestoolkit/rheachebiannotation/main.nf
mberacochea: I like MPT. I'll rename it.

Thoughts? React to this message with: 👍 or 👎  to vote

@KateSakharova @Ge94 @chrisAta @Vkale1 @ochkalova

modules/ebi-metagenomics/mgnifypipelinestoolkit/rheachebiannotation/main.nf
mberacochea: yup

modules/ebi-metagenomics/mgnifypipelinestoolkit/rheachebiannotation/main.nf
KateSakharova: I also think too much shorting name will cause confusion. Having it like Sonya suggested (mg_toolkit/mgtoolkit) or initial name seems more clear to me

mberacochea: We already use mg-tookit for the cli tool for our API https://pypi.org/project/mg-toolkit/, and there is another tool with that name 
mgtoolkit -> https://mgtoolkit.readthedocs.io/en/latest/installation.html is already "taken"... I want to avoid the confusion.

I think MgnifyPipelinesToolkit (MPT) is the best... or just keeping the long name (which is the convention). Let's see if we get a few more thoughts on Slack.

KateSakharova: I will vote for long name then :)

mberacochea: THE PEOPLE HAVE SPOKEN

modules/ebi-metagenomics/mgnifypipelinestoolkit/rheachebiannotation/meta.yml
mberacochea: Please correct it on the toolkit whenever you can :)

PR #77, opened 2025-02-06
dbcan modules

jmattock5:
Modules to download the dbCAN database and to run dbCAN on a protein input based off of those in mettannotator. 

Currently formatted to run on the pre-existing version of the dbCAN database on the ftp. Once that's been updated lines 24-26 in `modules/ebi-metagenomics/dbcan/dbcandb/main.nf` need deleting (I formatted the new db to have the correct name before compressing so it no longer needs renaming).

I've included a mini dbCAN database for the nf-test for the dbCAN module


modules/ebi-metagenomics/dbcan/dbcandb/main.nf
```diff
@@ -0,0 +1,48 @@
+process DBCANDB {
+    tag "dbCAN $params.db_version"
```
vagkaratzas, line 2: tag with meta.id, as the rest of the nf-core modules

modules/ebi-metagenomics/dbcan/dbcandb/main.nf
```diff
@@ -0,0 +1,48 @@
...
+    path "versions.yml"           , emit: versions
+
+    when:
+    task.ext.when == null || task.ext.when
+
+
+script:
+    def args = task.ext.args ?: ''
+    """
+    wget "${params.db_ftp_link}"
```
vagkaratzas, line 21: as we talked, define the db_version and/or url link in the def section above and use that groovy parameter in here instead

modules/ebi-metagenomics/dbcan/dbcandb/meta.yml
```diff
@@ -0,0 +1,26 @@
...
+  - CAZyme
+  - protein
+tools:
+  - "GNU Wget":
+      description: "Utility to download content from webs servers"
+      homepage: "https://www.gnu.org/software/wget/"
+      documentation: "https://www.gnu.org/software/wget/manual/"
+      tool_dev_url: "None"
+      licence: ["GPL v3"]
+output:
```
vagkaratzas, line 16: add inputs section..mostly copy paste from another nf-core module (dont forget the meta!)

modules/ebi-metagenomics/dbcan/dbcan/meta.yml
```diff
@@ -0,0 +1,46 @@
...
+input:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing sample information
+        e.g. `[ id:'sample1', single_end:false ]`
+  - faa:
+      type: file
+      description: Protein FASTA file containing query sequences
+      pattern: "*.{fasta,faa,fa}"
```
vagkaratzas, line 27: can it also work with .gz fasta files? if so add those extra extensions here

modules/ebi-metagenomics/dbcan/dbcan/main.nf
```diff
@@ -0,0 +1,51 @@
...
+    cat <<-END_VERSIONS > versions.yml
+    "${task.process}":
+        dbcan: "${params.dbcan_version}"
+        dbcan_db: "${params.db_version}"
+    END_VERSIONS
+    """
+
+    stub:
+    def args = task.ext.args ?: ''
+    def VERSION = "${params.dbcan_version}"
```
vagkaratzas, line 40: make sure there is no other way to get the dbcan version through the command line, before you hard code it here

modules/ebi-metagenomics/dbcan/dbcan/main.nf
```diff
@@ -0,0 +1,51 @@
+process DBCAN {
+    label 'process_medium'
```
vagkaratzas, line 2: tag with meta.id

modules/ebi-metagenomics/dbcan/dbcan/tests/main.nf.test
```diff
@@ -0,0 +1,45 @@
+nextflow_process {
+
+    name "Test Process DBCAN"
+    script "../main.nf"
+    process "DBCAN"
+
+    tag "modules"
+    tag "modules_nfcore"
+    tag "dbcan"
+
```
vagkaratzas, line 10: you are probably missing the setup step, where you run the dbcandb module first to create the db that dbcan then uses. see here for an example:t
https://github.com/nf-core/modules/blob/master/modules/nf-core/diamond/blastp/tests/main.nf.test
diamond/blastp first requires diamond/createdb to be run

vagkaratzas (changes requested):
Some starting comments here and there. dbcandb will also probably need its own little nf-test. Except if the database it downloads is huge I guess. Maybe if there is a smaller version?

modules/ebi-metagenomics/dbcan/dbcan/meta.yml
jmattock5: Checked and it does not

modules/ebi-metagenomics/dbcan/dbcan/main.nf
jmattock5: I did, annoyingly there is a version option but it just prints nothing

modules/ebi-metagenomics/dbcan/dbcan/tests/main.nf.test
jmattock5: As I said in another comment, the dbcandb module downloads the reference database, no other setup needed 😄

jmattock5:
I think the database is too big to download as a test but if @mberacochea agrees we could also upload the test db to the ftp and I could make a test with that.

Think I've addressed all of your comments Vangelis. Just seen that the nf-test has failed - made a new output file that it's never made before even though I've run it countless times 🤦‍♀️ will sort that next week.

modules/ebi-metagenomics/dbcan/dbcan/tests/main.nf.test
mberacochea: Yeah, the DB doesn't have to be built... it already is.

vagkaratzas (changes requested):
Some extra comments

mberacochea:
> I think the database is too big to download as a test but if @mberacochea agrees we could also upload the test db to the ftp and I could make a test with that.
> 
> Think I've addressed all of your comments Vangelis. Just seen that the nf-test has failed - made a new output file that it's never made before even though I've run it countless times 🤦‍♀️ will sort that next week.

You can also just add it here... which I think you already did

mberacochea:
@tgurbich can you please take a look at this one please?

modules/ebi-metagenomics/dbcan/dbcan/main.nf
```diff
@@ -0,0 +1,54 @@
...
+    script:
+    def args = task.ext.args ?: ''
+    """
+    run_dbcan \\
+        ${args} \\
+        --dia_cpu ${task.cpus} \\
+        --hmm_cpu ${task.cpus} \\
+        --tf_cpu ${task.cpus} \\
+        --dbcan_thread ${task.cpus} \\
+        --db_dir \$(pwd)/dbcan_db \\
```
vagkaratzas, line 30:
```suggestion
        --db_dir dbcan_db \\
```
The pwd is not needed here

modules/ebi-metagenomics/dbcan/dbcan/main.nf
```diff
@@ -0,0 +1,54 @@
...
+        --out_dir dbcan \\
+        --cgc_substrate \\
```
vagkaratzas, line 32:
```suggestion
```

modules/ebi-metagenomics/dbcan/dbcan/tests/main.nf.test
```diff
@@ -0,0 +1,105 @@
...
+                "substrate.out": "efca5ea50fef88b947ebdf89a34976a9",
+                "tp.out": "e9eb66ecff04b226deeb27a8fade736a",
+                "uniInput": "d6ccc30e774cac8b58fa4be247211d61"
+            ]
+            def emptyFiles = [
+                "tf-1.out",
+                "tf-2.out"
+            ]
+
+            file(process.out.dbcan_output.get(0).get(1)).eachFileRecurse { file ->
```
vagkaratzas, line 61: In a real nf-core module, all of these would still go in the snapshot. .exists()  .contains() .match() combinations
I'll let @mberacochea decide if this is good enough for our nf-modules.

modules/ebi-metagenomics/dbcan/dbcan/tests/main.nf.test
```diff
@@ -0,0 +1,105 @@
...
+                    file("${moduleDir}/tests/dbcan_test_db", type: 'dir', checkIfExists: true),
+                    "4.1.4_V13"
+                ]
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
```
vagkaratzas, line 101:
```suggestion
                { assert process.success },
                { assert snapshot(process.out).match() }
```
Probably removed by mistake

modules/ebi-metagenomics/dbcan/dbcan/main.nf
```diff
@@ -0,0 +1,54 @@
...
+        --cluster ${gff} \\
+        ${faa} \\
+        protein
```
vagkaratzas, line 35: If this is only ever going to be run with the `protein` mode, leave as it is.
Else, also (just as `--cgc_substrate`) provide this `protein` flag through ext.args (nextflow.config in the tests folder here, and modules.config in a pipeline)

vagkaratzas (changes requested):
Final Round!

modules/ebi-metagenomics/dbcan/dbcan/main.nf
tgurbich: `--cluster` is also an optional parameter

jmattock5: I think you need `--cluster` to be able to run `protein` so they'd have to be linked in the config.

Could we just rename this module to dbcanprot then copy&slightly modify the module if in the future we want to run it on one of the other settings?

tgurbich: I believe the `--cluster` flag is only required if you want dbcan to annotate clusters from proteins. It is also able to run on proteins without identifying clusters and then the cluster flag is not needed.

vagkaratzas: If the input is a file, that parameter still needs to be there in the module, and in non protein cases, you will have to input empty `[]` and then build the argument within the module (def before the script).

jmattock5: You're right Tania 😄 

Guess we need to decide if we'd ever **not** want to predict clusters and substrates.

modules/ebi-metagenomics/dbcan/dbcan/meta.yml
```diff
@@ -0,0 +1,183 @@
...
+  - - meta:
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. `[ id:'test', single_end:false ]`
+    - fasta:
+        type: file
+        description: A FASTA file containing query sequences, they can be proteins of
+          contigs
+        pattern: "*.{fasta,faa,fa}"
```
tgurbich, line 28:
```suggestion
        pattern: "*.{fasta,faa,fa,fna}"
```

vagkaratzas (approved):
epic

mberacochea:
> Just tested with my test workflow and it's all good 🥳
> 
> I've checked and running with an assembly, using the mode prok and args `-c cluster` instead of `--cluster ${gff}` completes successfully. We need to make sure we include this in the README when it's created

We need another test for this.. in the future

PR #78, opened 2025-02-10
FrageGeneScan RS - FGS but faster

mberacochea:


vagkaratzas (approved):
Why not straight to nf-core? Yikes!
The infernal test was probably triggered because of that lines deletion

mberacochea:
> Why not straight to nf-core? Yikes! The infernal test was probably triggered because of that lines deletion

We need this for the combined gene caller and the nf-core tool doesn't support swf with modules from different orgs. When that happens we can push this one to nf-core

modules/ebi-metagenomics/fraggenescanrs/meta.yml
```diff
@@ -0,0 +1,105 @@
...
+        description: |
+          Groovy Map containing sample information
+          e.g. `[ id:'test', single_end:false ]`
+    - fasta:
+        type: file
+        description: Database of sequences in FASTA format
+        pattern: "*.{fasta,fna,fasta.gz,fna.gz}"
+  - - training_file_name:
+        type: string
+        enum:
```
ochkalova, line 31: What is enum? List of possible names for a training file?

ochkalova (approved):
Good work, Martin 💃

modules/ebi-metagenomics/fraggenescanrs/main.nf
```diff
@@ -0,0 +1,59 @@
...
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    def catcommand = fasta.name.endsWith(".gz") ? "zcat" : "cat"
+    """
+    ${catcommand} ${fasta} | FragGeneScanRs \\
+        $args \\
```
KateSakharova, line 30: We need to remember about args in pipelines. Example, FGS had those https://github.com/EBI-Metagenomics/nf-modules/blob/main/modules/ebi-metagenomics/fraggenescan/main.nf#L29

command I tested:
`FragGeneScanRs -s genomic.fna -t illumina_10 -w 0 -o fgsR -p 4`
`-w` should be in args I think

modules/ebi-metagenomics/fraggenescanrs/meta.yml
```diff
@@ -0,0 +1,104 @@
...
+          - complete
+          - sanger_5
+          - sanger_10
+          - 4_5_4_5
+          - 4_54_10
+          - 4_54_30
```
KateSakharova, line 37: Those should be
454_5
454_10
454_30
```
    -t, --training-file <train_file_name>      File name that contains model parameters; this file should be in the -r
                                               directory or one of the following:
                                               [complete] for complete genomic sequences or short sequence reads without
                                               sequencing error
                                               [sanger_5] for Sanger sequencing reads with about 0.5% error rate
                                               [sanger_10] for Sanger sequencing reads with about 1% error rate
                                               [454_5] for 454 pyrosequencing reads with about 0.5% error rate
                                               [454_10] for 454 pyrosequencing reads with about 1% error rate
                                               [454_30] for 454 pyrosequencing reads with about 3% error rate
                                               [illumina_1] for Illumina sequencing reads with about 0.1% error rate
                                               [illumina_5] for Illumina sequencing reads with about 0.5% error rate
                                               [illumina_10] for Illumina sequencing reads with about 1% error rate
```

mberacochea: That was the linter `--fix`!

modules/ebi-metagenomics/fraggenescanrs/main.nf
mberacochea: Yes, the complete one is 0 by default:

```
-w, --complete <complete> The input sequence has complete genomic sequences; not short sequence reads. [default: 0]
```
The train "mode" is a mandatory arg -- I think having it there is better than in a config file as it's a mantdatory arg in this version of FGS.

KateSakharova (approved):

PR #80, opened 2025-02-11
Replace FGS with FGSRS in the CGC - and other tweaks

mberacochea:


modules/ebi-metagenomics/combinedgenecaller/merge/main.nf
```diff
@@ -7,7 +7,7 @@ process COMBINEDGENECALLER_MERGE {
         'biocontainers/mgnify-pipelines-toolkit:0.2.0--pyhdfd78af_0' }"
```
chrisAta, line 7: just FYI, the most recent version of the toolkit is 0.2.1 now

modules/ebi-metagenomics/combinedgenecaller/merge/main.nf
```diff
@@ -7,7 +7,7 @@ process COMBINEDGENECALLER_MERGE {
...
 
     input:
-    tuple val(meta), path(prodigal_sco, stageAs: "pro.sco"), path(prodigal_ffn, stageAs: "pro.ffn"), path(prodigal_faa, stageAs: "pro.faa"), path(fgs_out, stageAs: "fgf.out"), path(fgs_ffn, stageAs: "fgf.ffn"), path(fgs_faa, stageAs: "fgs.faa"), path(mask_file)
+    tuple val(meta), path(prodigal_sco, stageAs: "prodigal/"), path(prodigal_ffn, stageAs: "prodigal/"), path(prodigal_faa, stageAs: "prodigal/"), path(fgs_out, stageAs: "fgsrs/"), path(fgs_ffn, stageAs: "fgsrs/"), path(fgs_faa, stageAs: "fgsrs/"), path(mask)
```
chrisAta, line 10: is the `stageAs` just putting those files in respective subdirectories? e.g. for `prodigal_sco` it goes from `blabla.sco` to `prodigal/blabla.sco`? (making up names here haha)

modules/ebi-metagenomics/combinedgenecaller/merge/main.nf
```diff
@@ -56,9 +90,11 @@ process COMBINEDGENECALLER_MERGE {
     """
     touch ${prefix}.faa
     touch ${prefix}.ffn
+    touch ${prefix}.out
 
     gzip ${prefix}.faa
     gzip ${prefix}.ffn
+    gzip ${prefix}.out
```
chrisAta, line 97: why not do this in one line like you did in the main script
```suggestion
gzip -n ${prefix}.{faa,ffn,out}
```

subworkflows/ebi-metagenomics/combined_gene_caller/main.nf
```diff
@@ -17,8 +16,8 @@ workflow COMBINED_GENE_CALLER {
     PRODIGAL ( ch_assembly, channel.value("sco") )
     ch_versions = ch_versions.mix(PRODIGAL.out.versions)
 
-    FRAGGENESCAN ( ch_assembly )
-    ch_versions = ch_versions.mix(FRAGGENESCAN.out.versions)
+    FRAGGENESCANRS ( ch_assembly, "illumina_5" )
```
chrisAta, line 19: is this `illumina_5` always going to be consistent? could it change if we want to use the CGC on non-illumina data?

subworkflows/ebi-metagenomics/combined_gene_caller/tests/main.nf.test
```diff
@@ -12,14 +12,67 @@ nextflow_workflow {
     tag "prodigal"
     tag "fraggenescan"
 
-    test("it should work properly with a fasta as input") {
+    test("assembly fasta ") {
```
chrisAta, line 15: maybe write that it's a cgc test in the test name?

chrisAta (approved):
Looks good! just a few small comments, so I'll approve it

modules/ebi-metagenomics/combinedgenecaller/merge/main.nf
mberacochea: the problem is that is _very_ like that the files are called the same.. causing the process to fail.. os I place them in different folders to avoid that issue

subworkflows/ebi-metagenomics/combined_gene_caller/main.nf
mberacochea: That should be the default... I'll push that to a parameter (I forgot)

modules/ebi-metagenomics/combinedgenecaller/merge/main.nf
mberacochea: ahh .. perfect.

mberacochea:
There is one unit test  failling for deeptmhmm (which calls an API).. I'm ignore that one just now

PR #82, opened 2025-02-28
braker

vagkaratzas:
Needs to set `--AUGUSTUS_BIN_PATH` and  `--AUGUSTUS_SCRIPTS_PATH` from ext.args.
Sadly the conda environment doesn't come with preinstalled augustus stuff, I think.
So it is impossible atm to send directly to nf-core (or pass the tests I guess).
Should be OK to use in a pipeline though.

To port to nf-core, we would probably need to create an augustus nf-core module first, use it as a setup step and properly set these env variables before calling braker.

### DIFF
--- a/modules/ebi-metagenomics/braker/braker/environment.yml
+++ b/modules/ebi-metagenomics/braker/braker/environment.yml
@@ -1,0 +1,7 @@
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/environment-schema.json
+channels:
+  - conda-forge
+  - bioconda
+dependencies:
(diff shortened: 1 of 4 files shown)


vagkaratzas:
I remember this was usable if you would provide the AUGUSTUS stuff by hand. Could merge as is in the repo, if anyone wants to try this at some point in a pipeline. I believe that Varsha didn't need this after all.

mberacochea:
Well.. if you (or someone else) has the time to fix the nf-tests (or disable them with a DISCLAMER note) then merge it :)

vagkaratzas:
Or close the PR but please keep the branch for future reference

PR #83, opened 2025-03-04
Upgrade interproscan to version 5.73-104.0

mberacochea:
Removed the appls flag and run the tests will the defaul apps, I caught a few bugs with like this that I would have missed otherwise.

Remove the out format, it output all the 4 flavours. I think this is a better default that just one output kind, MGnify-ASA for example uses the TSV and GFF.

Also, I've pushed the "full" mini testing DB - which enables to run all the default apps with IPS.


mberacochea:
Cross PR - https://github.com/EBI-Metagenomics/containers/pull/33 (the container)

modules/ebi-metagenomics/interproscan/meta.yml
```diff
@@ -6,55 +6,73 @@ keywords:
...
   - "interproscan":
-      description: "InterPro integrates together predictive information about proteins function from a number of partner resources"
+      description: "InterPro integrates together predictive information about proteins
+        function from a number of partner resources"
       homepage: "https://www.ebi.ac.uk/interpro/search/sequence/"
       documentation: "https://interproscan-docs.readthedocs.io"
       tool_dev_url: "https://github.com/ebi-pf-team/interproscan"
       doi: "10.1093/bioinformatics/btu031"
       licence: ["GPL v3"]
+      identifier: ""
```
vagkaratzas, line 16:
```suggestion
      identifier: biotools:interproscan_ebi
```

modules/ebi-metagenomics/interproscan/meta.yml
```diff
@@ -6,55 +6,73 @@ keywords:
...
+        description: Version of the interproscan database for the versions.yml
 output:
   - tsv:
-      type: file
-      description: Tab separated file containing with detailed hits
-      pattern: "*.{tsv}"
+      - meta:
+          type: file
+          description: Tab separated file containing with detailed hits
+          pattern: "*.{tsv}"
```
vagkaratzas, line 39: metas shouldn't have a pattern.
Same for the rest below

modules/ebi-metagenomics/interproscan/meta.yml
```diff
@@ -6,55 +6,73 @@ keywords:
...
+      - "*.tsv.gz":
+          type: file
+          description: Tab separated file containing with detailed hits
+          pattern: "*.{tsv}"
```
vagkaratzas, line 43: patterns should also include the .gz extension.
Same for the rest below

vagkaratzas (changes requested):
Nice! Without `-appl `, is it running for every folder in /data?
Did you make small subsets for each and every one of the dbs? o.O

modules/ebi-metagenomics/interproscan/meta.yml
mberacochea: yup.. I just ran nf-core lint -fix and the linter was happy. Good catch

mberacochea:
> Nice! Without `-appl `, is it running for every folder in /data? Did you make small subsets for each and every one of the dbs? o.O

yeah, it's running the whoel thing. The IPS folks added a mini db to test things, I had to compress some HMM last time... but they fixed that now.

modules/ebi-metagenomics/interproscan/meta.yml
mberacochea: good catch!

vagkaratzas (approved):
LGTM!

mberacochea:
The unit test is failling because it can't find `ps` in the container.. but I works locally. I think I pushed a faulty container and it was cached. I'll merge this one and will test it in the cluster.

mberacochea:
I can't get the singularity version to work, I'll merge and debug this in the cluster as my 🍎 M4 can't run singularity

PR #84, opened 2025-03-07
Assembly decontamination subworkflow

jmattock5:
Subworkflow that takes an assembly and blast database as input, blasts against the ref and removes host contamination. Will need to be run twice if removing human+phiX and another host.

Uses the same blast parameters as miassembler with the addition of a query coverage threshold of 50%.

The modules added are the current versions of blast and seqkit grep from nf-core. FYI this version of the blast module uses meta as the output file prefix rather than meta2 so the output is now sample1.txt rather than chicken.txt. Easy to change back if needed.


jmattock5:
The pre-commit failure is because prettier doesn't like the format of one of the test blast db files, we can't edit it so can we get pre-commit to ignore that file?

The nf-test failure is the nf-core blast module failing, this was downloaded from the nf-core modules GitHub repo and I haven't edited a thing in it.

subworkflows/ebi-metagenomics/assembly_decontamination/tests/main.nf.test
```diff
@@ -0,0 +1,36 @@
...
+                    file("${moduleDir}/tests/data/test.fa", checkIfExists: true)
+                ])
+                input[1] = "contamination.fa"
+                """
+            }
+        }
+        then {
+            assertAll(
+                { assert workflow.success },
+                { assert snapshot(workflow.out.cleaned_contigs).match() }
```
mberacochea, line 32: I think it's worth it to check which contigs are being returned here. There is a plugin for nf-test that allows to do assertions on fasta files (https://www.nf-test.com/docs/assertions/fasta/)

I think in your case it would look like:

```
with(workflow.out.cleaned_contigs) {
  def clean_contigs = path(get(0)).fasta
  assert "uncontaminated_contig" in clean_contigs
  assert !("partially_contaminated_contig" in sequences)
  assert !("contaminated_contig" in sequences)
}
```

mberacochea (changes requested):
Thank you @jmattock5 . It looks pretty good to me, I just left a couple of things to address before we merget this one.

subworkflows/ebi-metagenomics/assembly_decontamination/main.nf
```diff
@@ -0,0 +1,27 @@
...
+    assembly               // [ val(meta), path(assembly_fasta) ]
+    reference_genome       // [ path(reference_genome) ]
+
+    main:
+    ch_versions = Channel.empty()
+    cleaned_contigs = Channel.empty()
+    ch_blast_ref = Channel.fromPath( "${params.blast_reference_genomes_folder}/${reference_genome}*", checkIfExists: true)
+            .collect().map {
+                files -> [ ["id": reference_genome], files ]
+            }
```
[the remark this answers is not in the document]
jmattock5: Do you mean changing it to this:
```
    ch_blast_ref = Channel.fromPath( "${reference_genome}*", checkIfExists: true)
            .collect().map {
                files -> [ ["id": reference_genome], files ]
            }
```

jmattock5:
Thanks @mberacochea, I've made all the changes you suggested

subworkflows/ebi-metagenomics/assembly_decontamination/main.nf
```diff
@@ -0,0 +1,22 @@
...
+    assembly           // [ val(meta), path(assembly_fasta) ]
+    ch_blast_ref       // [ val(meta2), path(reference_genome_files) ]
+
+    main:
+    ch_versions = Channel.empty()
+
+    BLAST_BLASTN(assembly,ch_blast_ref)
+    ch_versions = ch_versions.mix(BLAST_BLASTN.out.versions.first())
+
+    SEQKIT_GREP(assembly,BLAST_BLASTN.out.txt.flatten().last())
```
mberacochea, line 16: I missed this the first time. Why are you using `flatten().last()` here?

subworkflows/ebi-metagenomics/assembly_decontamination/tests/main.nf.test
```diff
@@ -0,0 +1,43 @@
...
+        }
+        then {
+            assertAll(
+                { assert workflow.success },
+                { assert snapshot(workflow.out.cleaned_contigs).match() },
+                {
+                  def cleaned_contigs = path(workflow.out.cleaned_contigs.get(0).get(1)).fasta
+                  assert "uncontaminated_contig" in cleaned_contigs
+                  assert "partially_contaminated_contig" in cleaned_contigs
+                  assert !("contaminated_contig" in cleaned_contigs)
```
mberacochea, line 38: Excellent!

mberacochea (commented):
Cool... there is one test failling (blast makedb?).

I left a question

subworkflows/ebi-metagenomics/assembly_decontamination/main.nf
jmattock5: BLAST_BLASTN.out.txt is in the format `val(meta), path('*.txt')` and seqkit_grep kept erroring trying to use val(meta) as the path. So I split the channel with flatten and used last to get the path. Please change if there's a better way to do this.

mberacochea: ahh.. you should use it like this then:
```
SEQKIT_GREP(assembly,BLAST_BLASTN.out.txt.map { meta, txt -> txt }.first() )
```

the `map` operator will run a function to each value on the channel. In our case it just returns the `txt` path. I'm using `first()` to convert the channel to a file thingy (this is a nextflow obscure thing).. if you don't do this and you have more than one assembly it will only run once as the txt channel has only one element

jmattock5: Have changed it as suggested 😄

mberacochea (approved):
Good stuff. Thank you Jenny

PR #87, opened 2025-03-20
Detect rna general subwf

KateSakharova:


mberacochea:
I've fixed the linting and unit tests issues, I merged this to test it with https://github.com/EBI-Metagenomics/assembly-analysis-pipeline.

But, there are a few things to do:
- Add tests with ncRNA, as the current test in extractcoords only cover rRNA
- Review the outputs of extractcoords

`extractcoords` as a module name it's not very good (we should look into renaming it in the future)

PR #88, opened 2025-03-24
Update the CONTIGS_TAXONOMIC_CLASSIFICATION

ochkalova:
Subworkflow updated to produce KRONA plot
Tests updated

WIP:
add tags to tests
currently local module is used instead of mgnify-toolkit, I need new release of mgnify-toolkit to update it


mberacochea (approved):
Thank you @ochkalova !

PR #95, opened 2025-05-20
Update the IPS module

mberacochea:
- Handle .gz inputs
- Add licensed software support (containers only)


modules/ebi-metagenomics/interproscan/main.nf
```diff
@@ -29,11 +37,17 @@ process INTERPROSCAN {
     script:
     def args = task.ext.args ?: ''
     def prefix = task.ext.prefix ?: "${meta.id}"
-    // -dp (disable precalculation) is on so no online dependency
+    def is_compressed = fasta.getExtension() == "gz"
+    def fasta_file_name = fasta.name.replace(".gz", "")
```
vagkaratzas, line 41: This is a bit clunky...if for some reason there is ".gz" inside the filename, anywhere else than just at the end :(
Can you just remove from the extension?

vagkaratzas (approved):
Approved but have a look into that .gz extension removal thingie. Sure we've done it for multiple nf-core modules

modules/ebi-metagenomics/interproscan/main.nf
mberacochea: I would argue if you have a file named "thing.gz.gz" then that is an issue. I'll change this, a groovy regex clean should be simple enough.

```
def fasta_file_name = fasta.name - ~/\.gz$/
```
this should only remove the trailling ".gz".

PR #98, opened 2025-06-17
Update readsqc swf pimento 1.0.1

chrisAta:
The READS_QC subworkflow uses a script from the mgnify-pipelines-toolkit for that helps identify whether the library strategy is amplicon or not (librarystrategycheck) that takes as input the output of a now deprecated script from the toolkit (assessmcpproportions). The equivalent script of that deprecated script now exists instead in [PIMENTO](https://github.com/EBI-Metagenomics/PIMENTO/blob/dev/pimento/bin/generate_bcv.py), so this PR updates the subworkflow to use that in a new module

### DIFF
--- a/modules/ebi-metagenomics/librarystrategycheck/tests/main.nf.test
+++ b/modules/ebi-metagenomics/librarystrategycheck/tests/main.nf.test
@@ -25,7 +25,11 @@ nextflow_process {
         then {
             assertAll(
                 { assert process.success },
-                { assert snapshot(process.out).match() }
+                { assert snapshot(
+                    process.out.library_check_out,
+                    process.out.versions.collect { path(it).yaml }
+                    ).match()
+                }
             )
         }
     }
@@ -46,7 +50,11 @@ nextflow_process {
         then {
             assertAll(
                 { assert process.success },
-                { assert snapshot(process.out).match() }
+                { assert snapshot(
+                    process.out.library_check_out,
+                    process.out.versions.collect { path(it).yaml }
+                    ).match()
+                }
             )
         }
     }
@@ -69,7 +77,11 @@ nextflow_process {
         then {
             assertAll(
                 { assert process.success },
-                { assert snapshot(process.out).match() }
+                { assert snapshot(
+                    process.out.library_check_out,
+                    process.out.versions.collect { path(it).yaml }
+                    ).match()
+                }
             )
         }
     }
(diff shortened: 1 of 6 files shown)


modules/ebi-metagenomics/pimento/generatebcv/meta.yml
```diff
@@ -37,10 +37,7 @@ input:
         type: file
         description: Merged or single-end FASTQ file
         pattern: "*.fastq.gz"
-  - - library_check:
```
vagkaratzas, line 40: missing `fwd_flag` etc

modules/ebi-metagenomics/pimento/generatebcv/tests/main.nf.test
```diff
@@ -20,7 +21,6 @@ nextflow_process {
                     "auto",
```
vagkaratzas, line 21: I don't think this format matches this:
`tuple val(meta), val(fwd_flag), val(rev_flag), path(fastq)`
Am I misreading something?

modules/ebi-metagenomics/librarystrategycheck/tests/main.nf.test
```diff
@@ -0,0 +1,76 @@
...
+                    file("${moduleDir}/tests/fixtures/test_bcv.tsv", checkIfExists: true)
+                    ]
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert process.success },
+                { assert snapshot(process.out).match() }
```
vagkaratzas, line 28:
```suggestion
                { assert snapshot(
                    process.out.library_check_out,
                    process.out.versions.collect { path(it).yaml }
                    ).match()}
```
A nicer way to print actual versions once in the snapshot nowadays, instead of just a hash

vagkaratzas (changes requested):
Nice, could be ported to nf-core probably with minor tweaks. Some comments here and there for now

modules/ebi-metagenomics/pimento/generatebcv/meta.yml
chrisAta: nope they're here! scroll up :D

modules/ebi-metagenomics/pimento/generatebcv/tests/main.nf.test
chrisAta: ah nope nothing wrong with this, that is indeed a tuple with meta, val, val, and then a path

vagkaratzas (approved):
Good!

PR #105, opened 2025-09-10
Add --threads to proovfread map

mberacochea:


KateSakharova (changes requested):
We can adjust only `args` in config. Why do you need it in module? The same for diamond mode. I just modified those in modules.config in pipeline

PR #109, opened 2025-09-22
Modules and subworkflows documentation website - built with MKDocs

mberacochea:
This PR introduces a documentation bit using MkDocs to automatically generate them for all modules and subworkflows in the repository. It uses MKDocs Material (https://squidfunk.github.io/mkdocs-material/), which provides loads of goodies on top of MkDocs.
There is a python script that reads the meta.yml of modules and subworkflows and using jinja2 tempaltes generates .md for MkDocs. 
The website is hosted in github pages, and automatically deployed :).

Website: https://ebi-metagenomics.github.io/nf-modules/modules

The index page doesn't work for some reason, it's throwing a 404 (I'll fix it before merging).

Claude code did most of the scripting, but I had to adjust it to make it work the way I wanted.

### DIFF
--- a/docs/templates/module.md.j2
+++ b/docs/templates/module.md.j2
@@ -1,6 +1,7 @@
 ---
 hide:
   - navigation
+title: {{ meta.name }}
 ---
 
 ## Description
--- a/docs/templates/subworkflow.md.j2
+++ b/docs/templates/subworkflow.md.j2
@@ -1,6 +1,7 @@
 ---
 hide:
   - navigation
+title: {{ meta.name }}
 ---
 
 ## Description


SandyRogers (approved):
Lovely stuff, thanks @mberacochea .

My only big query/suggestion is to consider (if you didn't already) using the nf-core python package to simplify `generate_docs` a bit?

E.g. instead of using the local file system to find `meta.yml`s, we could always publish docs from the committed modules I suppose, something like
```python
from nf_core.modules.info import ModuleInfo
info=ModuleInfo(pipeline_dir="ebi-metagenomics", component_name="antismash", remote_url="https://github.com/ebi-metagenomics/nf-modules")
info.get_remote_yml()
```

Another thing – not sure if valuable to the nf-core community or not though – would be to add a method to that ModuleInfo like `generate_markdown_doc()` or something.

Neither of these are important/blockers though. Nice.

PR #111, opened 2025-10-02
Update EXTRACTCOORDS module

chrisAta:
Things were a bit of out sync with the EXTRACTCOORDS module in a couple places (amplicon/assembly analysis pipelines), updating it in nf-modules so that we can reset the status of this module again

I'm also unsure whether `--separate-subunits-by-models` should be a default flag or not. It will be used in the amplicon-pipeline but not sure about assembly


subworkflows/ebi-metagenomics/rrna_extraction/main.nf
```diff
@@ -37,7 +37,8 @@ workflow RRNA_EXTRACTION {
 
     EXTRACTCOORDS(
         EASEL_ESLSFETCH.out.easel_coords,
-        EASEL_ESLSFETCH.out.matched_seqs_with_coords
+        EASEL_ESLSFETCH.out.matched_seqs_with_coords,
+        []
```
mberacochea, line 41:
```suggestion
        false
```

PR #122, opened 2025-10-21
Modify sam flag in bwamem2decontnobams for paired reads

timrozday-mgnify:
`-f 4` keeps only unmapped reads but ignores the pair (read1 unmapped)
`-f 12` keeps reads only if both it and its pair are unmapped (read1 unmapped + read2 unmapped)

For paired reads we want to remove them if either of the reads in the pair map to the decontamination genome, so we need to change the flag to `-f 12`.

### DIFF
--- a/modules/ebi-metagenomics/bwamem2decontnobams/main.nf
+++ b/modules/ebi-metagenomics/bwamem2decontnobams/main.nf
@@ -42,7 +42,7 @@ process BWAMEM2DECONTNOBAMS {
             -t $task.cpus \\
             \$INDEX \\
             $reads \\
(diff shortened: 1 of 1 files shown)


Ales-ibt (approved):
It makes sense. Thank you.

jmattock5:
But we currently use the reads decontaminated by this module in the coverage calculation subworkflow in miassembler

jmattock5:
But if the unmapped read is contamination then it should be mapping to the ref too. One explanation is it could be a chimera, then we may be discarding microbial DNA that may not assemble but could give useful coverage info if it mapped elsewhere

PR #148, opened 2026-03-20
gff2gbk/ module

Ales-ibt:
This module runs with mgnify pipelines toolkit. It generates gbk files from proteins and gff files. This is part of the BGCs integrator subworkflow.

Thanks for reviewing!


vagkaratzas (approved):
While I understand that the mgnify toolkit contains mostly helper functions for within the group consumption, there are some commands that would be really helpful for the community, like this one here. I would propose exposing a named API for the library, to have commands such as `mgnifytoolkit gff2gbk --input ... --output ...` and start pushing modules to nf-core. For example this would be:
`MGNIFYTOOLKIT_GFF2GBK`
It is already on bioconda and PyPI so it would be piece of cake. I would just spend some time optimizing all python commands along with the bots. Tagging @mberacochea to discuss

mberacochea (changes requested):
We should have these in the mgnifypipelinestoolkit folder

mberacochea (approved):
@Ales-ibt 🥇

PR #149, opened 2026-03-20
bgcmapper module for the bgcs integrator subwf

Ales-ibt:
This is the second module that we need for the bgc integrator subworkflow. The code in oython is already in the mgnify pipelines toolkit. 

Thanks for reviewing!


vagkaratzas (approved):
LGTM!

mberacochea (changes requested):
I rather have all the toolkit modules in the same subworkflow - https://github.com/EBI-Metagenomics/nf-modules/tree/main/modules/ebi-metagenomics/mgnifypipelinestoolkit

it makes more logical sense (more organized) and easier to auto-update in the future.

mberacochea (approved):
Thanks @Ales-ibt !

PR #150, opened 2026-03-25
bgcs subwf

Ales-ibt:
This is (finally!!!!) the BGCs subworkflow in place.
I removed the downloading of the IPS database because it was adding too much unnecessary complexity to the code and debugging it was driving me crazy.

Thanks for reviewing!


subworkflows/ebi-metagenomics/bgc_annotation/tests/nextflow.config
```diff
@@ -0,0 +1,16 @@
...
+    withName: BGCSMAPPER {
+        ext.args = "--validate_json"
+    }
+    withName: 'BGC_ANNOTATION:GECCO_RUN' {
+        container = 'community.wave.seqera.io/library/gecco:0.10.1--35736d55c7f3d788'
+        containerOptions = '--platform linux/arm64'
+    }
+    withName: 'BGC_ANNOTATION:GECCO_CONVERT' {
+        container = 'community.wave.seqera.io/library/gecco:0.10.1--35736d55c7f3d788'
+        containerOptions = '--platform linux/arm64'
```
mberacochea: the platform is specific to where the pipeline is running, so this is managed by the profiles (usually).

For example, most of our pipeline have an `arm` profile that set this flag - 
```
    arm {
        docker.runOptions = '-u $(id -u):$(id -g) --platform=linux/amd64'
    }
```

subworkflows/ebi-metagenomics/bgc_annotation/main.nf
```diff
@@ -0,0 +1,192 @@
...
+    GFF2GBK(ch_togbk_input)
+
+    // Run SanntiS
+    if (!skip_sanntis) {
+
+        // Split IPS input into provided vs missing
+        ch_ips.branch { meta, ips_annot ->
+            provided: ips_annot != null
+            missing:  ips_annot == null
+        }.set { ch_ips_branched }
```
mberacochea:
```suggestion
        ch_ips_branched = ch_ips.branch { meta, ips_annot ->
            provided: ips_annot != null
            missing:  ips_annot == null
        }
```
the `.set` thingy will eventually go away (because the Nextflow lang parser struggles with it) so better stay ahead of the curve

subworkflows/ebi-metagenomics/bgc_annotation/main.nf
```diff
@@ -0,0 +1,192 @@
...
+        // Samples missing IPS: join with proteins to feed InterProScan
+        ch_prots_missing = ch_prots
+            .join(ch_ips_branched.missing, by: 0)
+            .map { meta, prots, _ips_missing -> tuple(meta, prots) }
+
+        /*
+         * No automatic DB download anymore.
+         * If IPS is missing and INTERPROSCAN is needed, ch_ips_db must be provided.
+         */
+        INTERPROSCAN(ch_prots_missing, ch_ips_db)
```
mberacochea, line 57: I would would add some defensive code there

```suggestion
        ch_ips_db.ifEmpty { error("The InterProScan database not found. Automatic download is no longer supported. Please provide it via 'ch_ips_db'.") }

        INTERPROSCAN(ch_prots_missing, ch_ips_db)
```

subworkflows/ebi-metagenomics/bgc_annotation/main.nf
```diff
@@ -0,0 +1,192 @@
...
+
+        SANNTIS(ch_sanntis_input)
+        ch_versions = ch_versions.mix(SANNTIS.out.versions)
+        ch_sanntis_results = SANNTIS.out.gff
+    }
+
+    // Run GECCO
+    if (!skip_gecco) {
+        ch_gecco_input =  GFF2GBK.out.gbk.map { meta, gbk -> tuple(meta, gbk, []) }
+        GECCO_RUN(ch_gecco_input, [])
```
mberacochea, line 82: The versions thingy is missing for GECCO and GECCO_CONVERT

subworkflows/ebi-metagenomics/bgc_annotation/main.nf
```diff
@@ -0,0 +1,192 @@
...
+        ANTISMASH_ANTISMASHDOWNLOADDATABASES()
+        ch_versions = ch_versions.mix(ANTISMASH_ANTISMASHDOWNLOADDATABASES.out.versions)
+        antismash_db = ANTISMASH_ANTISMASHDOWNLOADDATABASES.out.database
+    }
+
+    if (!skip_antismash) {
+        ANTISMASH_ANTISMASH(GFF2GBK.out.gbk, antismash_db, [])
+        ch_versions = ch_versions.mix(ANTISMASH_ANTISMASH.out.versions)
+
+        ANTISMASH_JSON2GFF(ANTISMASH_ANTISMASH.out.json_results)
```
mberacochea, line 105: The versions thingy is missing for ANTISMASH_JSON2GFF

subworkflows/ebi-metagenomics/bgc_annotation/main.nf
```diff
@@ -0,0 +1,192 @@
...
+    ch_antismash_db     // channel: path(antismash_db)
+    ch_ips_db           // channel: path(interproscan_db)
+    skip_sanntis        // boolean
+    skip_gecco          // boolean
+    skip_antismash      // boolean
+
+    main:
+    ch_versions = Channel.empty()
+
+    // Extract individual inputs from input channel
```
KateSakharova, line 26: maybe you can use multiMap
```
def inputs = ch_inputs.multiMap { meta, contigs, gff, proteins, ips_annot ->
    gff:         tuple(meta, gff)
    togbk_input: tuple(meta, contigs, gff, proteins)
    ips:         tuple(meta, ips_annot)
    prots:       tuple(meta, proteins)
}
```
and then use directly in modules instead of `ch_gff` use `inputs.gff` etc
```
ch_gff         = inputs.gff
ch_togbk_input = inputs.togbk_input
ch_ips         = inputs.ips
ch_prots       = inputs.prots
```

subworkflows/ebi-metagenomics/bgc_annotation/main.nf
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    // Run GECCO
+    if (!skip_gecco) {
+        ch_gecco_input =  GFF2GBK.out.gbk.map { meta, gbk -> tuple(meta, gbk, []) }
+        GECCO_RUN(ch_gecco_input, [])
+
+        ch_gecco_convert_input = GECCO_RUN.out.clusters
+            .join(GECCO_RUN.out.gbk, by: 0)
+            .map { meta, clusters, gbk -> tuple(meta, clusters, gbk) }
```
KateSakharova: why do you need that map here? I think join is fine

subworkflows/ebi-metagenomics/bgc_annotation/main.nf
KateSakharova, line 57: I guess there should be a check that DB exists or provided. You are more sensitive for antismash DB. At least error message would be good!

subworkflows/ebi-metagenomics/bgc_annotation/main.nf
```diff
@@ -0,0 +1,192 @@
...
+
+    if (!skip_antismash) {
+        ANTISMASH_ANTISMASH(GFF2GBK.out.gbk, antismash_db, [])
+        ch_versions = ch_versions.mix(ANTISMASH_ANTISMASH.out.versions)
+
+        ANTISMASH_JSON2GFF(ANTISMASH_ANTISMASH.out.json_results)
+        ch_antismash_results = ANTISMASH_JSON2GFF.out.gff
+    }
+
+    // Ensure each optional result channel exists even if tool is skipped
```
mberacochea, line 109: I think the optional results handling can be simplified by:
- having ch_sanntis_results = channel.empty() before each if conditional (define as empty -- override if there are results
- then doing a single join, filter and map:
```
ch_for_bgcsmapper = ch_gff
    .join(ch_sanntis_results,   by: 0, remainder: true)
    .join(ch_gecco_results,     by: 0, remainder: true)
    .join(ch_antismash_results, by: 0, remainder: true)
    // Get rid of any samples with no annotations.. maybe we should fail here?
    .filter { meta, gff, sanntis_gff, gecco_gff, antismash_gff ->
        sanntis_gff || gecco_gff || antismash_gff
    }
    .map { meta, gff, sanntis_gff, gecco_gff, antismash_gff ->
        tuple(
            meta,
            gff,
            sanntis_gff   ?: [],
            gecco_gff     ?: [],
            antismash_gff ?: []
        )
    }
```

mberacochea (changes requested):
Great stuff @Ales-ibt, massive effort and the code very nice.

I left some notes, I think we can simplify a few things

subworkflows/ebi-metagenomics/bgc_annotation/tests/main.nf.test
```diff
@@ -0,0 +1,182 @@
...
+        when {
+            workflow {
+                """
+                input[0] = channel.of( [ [ id:'test' ], // meta map
+                    file('${moduleDir}/tests/data/test.fasta.gz', checkIfExists: true),
+                    file('${moduleDir}/tests/data/test.gff.gz', checkIfExists: true),
+                    file('${moduleDir}/tests/data/test.faa.gz', checkIfExists: true),
+                    file('${moduleDir}/tests/data/test_ips.tsv.gz', checkIfExists: true),
+                ])
+                input[1] = Channel.value(null)      // ch_antismash_db
```
mberacochea: Is then the workflow downloading the whole DB? We should use a dummy test one.. otherwise it will take ages

mberacochea (dismissed):

subworkflows/ebi-metagenomics/bgc_annotation/main.nf
```diff
@@ -0,0 +1,192 @@
...
+            tuple(
+                meta,
+                gff,
+                sanntis_gff  ?: [],
+                gecco_gff    ?: [],
+                antismash_gff ?: []
+            )
+        }
+
+    BGCSMAPPER(ch_for_bgcsmapper)
```
KateSakharova, line 186: From line 105-183 seems so complecated. I understood that it might be easier to debug
I have asked SeqeraAI and it suggested instead of 80 lines

Up to you to try it :)

```
// Find samples with at least one BGC result
ch_has_any_bgc = Channel.empty()
    .mix(skip_sanntis   ? Channel.empty() : ch_sanntis_results)
    .mix(skip_gecco     ? Channel.empty() : ch_gecco_results)
    .mix(skip_antismash ? Channel.empty() : ch_antismash_results)
    .map { meta, _file -> tuple(meta, true) }
    .unique()

// Start with filtered GFF
ch_for_bgcsmapper = ch_gff
    .join(ch_has_any_bgc)
    .map { meta, gff, _flag -> tuple(meta, gff) }
    // Join all tool results with remainder:true to keep all samples
    .join(skip_sanntis   ? Channel.empty() : ch_sanntis_results,   by: 0, remainder: true)
    .join(skip_gecco     ? Channel.empty() : ch_gecco_results,     by: 0, remainder: true)
    .join(skip_antismash ? Channel.empty() : ch_antismash_results, by: 0, remainder: true)
    // Replace nulls with []
    .map { meta, gff, sanntis, gecco, antismash ->
        tuple(meta, gff, sanntis ?: [], gecco ?: [], antismash ?: [])
    }

BGCSMAPPER(ch_for_bgcsmapper)
```

In seqsubmit we put some things into meta. Like meta.gecco = true, meta.sanntis=false. And used it for filtering and joining...

KateSakharova (changes requested):
I have added comments about nextflow language itself.

Could you also add comments to all `[]` because sometimes it is not clear what are those args for some modules (there are about 4-5 places).

Btw, I like those workflows! This one will become the same great as AMR detection ⭐

subworkflows/ebi-metagenomics/bgc_annotation/main.nf
Ales-ibt: topics :D

subworkflows/ebi-metagenomics/bgc_annotation/main.nf
Ales-ibt: Yes, when versions are defined in topics, we don't need to make the mix thingy

mberacochea: ah.. @vagkaratzas how do we handle topics and non-topics then? We need to do a sweep an update all the modules (not now BTW)

KateSakharova: Topics are automatically processed in ~dumversions module from nf-core as far as I know. For tools without topics - old way, collect with versions.yml

vagkaratzas: We ignore it in the subworkflows, exactly as Ales did. They will only be picked at the end of the pipeline.
Subworkflows will only have to output the versions, if at least one of its modules is still using the old way.

mberacochea: We need to be mindful when using this one then, I don't expect all our pipelines to be topics-ready

mberacochea: image
then.. thank you folks

subworkflows/ebi-metagenomics/bgc_annotation/main.nf
```diff
@@ -0,0 +1,192 @@
...
+    take:
+    ch_inputs           // channel: tuple( val(meta), path(contigs), path(gff), path(proteins), path(ips_annot) )
+    ch_antismash_db     // channel: path(antismash_db)
+    ch_ips_db           // channel: path(interproscan_db)
+    skip_sanntis        // boolean
+    skip_gecco          // boolean
+    skip_antismash      // boolean
+
+    main:
+    ch_versions = Channel.empty()
```
vagkaratzas:
```suggestion
    ch_versions = channel.empty()
```

subworkflows/ebi-metagenomics/bgc_annotation/main.nf
```diff
@@ -0,0 +1,192 @@
...
+        ch_versions = ch_versions.mix(ANTISMASH_ANTISMASH.out.versions)
+
+        ANTISMASH_JSON2GFF(ANTISMASH_ANTISMASH.out.json_results)
+        ch_antismash_results = ANTISMASH_JSON2GFF.out.gff
+    }
+
+    // Ensure each optional result channel exists even if tool is skipped
+    ch_sanntis_results   = skip_sanntis   ? Channel.empty() : ch_sanntis_results
+    ch_gecco_results     = skip_gecco     ? Channel.empty() : ch_gecco_results
+    ch_antismash_results = skip_antismash ? Channel.empty() : ch_antismash_results
```
vagkaratzas:
```suggestion
    ch_sanntis_results   = skip_sanntis   ? channel.empty() : ch_sanntis_results
    ch_gecco_results     = skip_gecco     ? channel.empty() : ch_gecco_results
    ch_antismash_results = skip_antismash ? channel.empty() : ch_antismash_results
```

subworkflows/ebi-metagenomics/bgc_annotation/tests/main.nf.test
```diff
@@ -0,0 +1,182 @@
...
+            }
+        }
+
+        then {
+            assert workflow.success
+            with(workflow.out.bgc_output) {
+                assert size() == 1
+                assert get(0)[0].id == "test"
+                assert get(0)[1].endsWith("_bgcs.gff")
+                assert get(0)[2].endsWith("_bgcs.json")
```
vagkaratzas, line 64: no snapshot? :'D

mberacochea:
@Ales-ibt I had to kill https://github.com/EBI-Metagenomics/nf-modules/actions/runs/23538203255 is was taking ages (> 2 hours). I suspect it is downloading the antiSMASH db which is massive

vagkaratzas (changes requested):
Nice work!! Whatever was already mentioned +
(i) change all remaining `Channel` to `channel`
(ii) add snapshots to nf-tests

subworkflows/ebi-metagenomics/bgc_annotation/tests/main.nf.test
Ales-ibt: I will fix the test to use https://github.com/EBI-Metagenomics/nf-modules/tree/main/modules/ebi-metagenomics/antismash/antismash/tests/antismash_trimmed

subworkflows/ebi-metagenomics/bgc_annotation/tests/main.nf.test
Ales-ibt: Not for the json because it has a timestamp. We test the json generation in the corresponding module https://github.com/EBI-Metagenomics/nf-modules/blob/0da36447750e5f12e61e2dd629b749f31f2c258f/modules/ebi-metagenomics/mgnifypipelinestoolkit/bgcsmapper/tests/main.nf.test#L33-L55

subworkflows/ebi-metagenomics/bgc_annotation/main.nf
Ales-ibt: I was trying to be sure about the order of the files in the new tuple, but I will remove it :D

subworkflows/ebi-metagenomics/bgc_annotation/main.nf
Ales-ibt: Thanks, same comment as Martin. I had committed a defensive code he suggested.

subworkflows/ebi-metagenomics/bgc_annotation/tests/main.nf.test
Ales-ibt: I will add a snapshot for the gff :D

subworkflows/ebi-metagenomics/bgc_annotation/main.nf
Ales-ibt: I doubled check, gecco run uses topics, gecco convert emits versions.yml. I made the addition

subworkflows/ebi-metagenomics/bgc_annotation/main.nf
Ales-ibt: Done

subworkflows/ebi-metagenomics/bgc_annotation/tests/main.nf.test
vagkaratzas: All of the assertions can go into a snapshot, even if they are not md5sum checks :D
My favourites in order:
md5sum -> number of lines (or even printing a couples of similar expected lines) -> filename (worst case)

subworkflows/ebi-metagenomics/bgc_annotation/main.nf
Ales-ibt: This is so nice!

Ales-ibt: Added!

subworkflows/ebi-metagenomics/bgc_annotation/main.nf
Ales-ibt: I removed this line because it was causing the process to hang when IPS do not need to run. I will just let the process to naturally fail and the user can find out why exploring the log files

subworkflows/ebi-metagenomics/bgc_annotation/tests/main.nf.test
Ales-ibt: I copied the minidatabase from the assembly analysis pipeline and it works :D

subworkflows/ebi-metagenomics/bgc_annotation/tests/main.nf.test
Ales-ibt: I have added assertions into snapshot for all the positive tests

Ales-ibt:
Thanks all of you for your reviews, addressing your comments definitely leads to a substancial improvement in the code quality. These are the main changes I made:

1. Fix tests adding snapshots and using a minidatabase for antismash
2. Add arm profile
3. I tried to make the last part of the code simpler but I constantly bumped into the issue of sanntis test hanging due to channel consumption issues, so I ended keeping the current old reliable code :D

As all of you asked for changes, I'd need you to approval for merging.

Thanks again!

subworkflows/ebi-metagenomics/bgc_annotation/tests/main.nf.test
```diff
@@ -0,0 +1,314 @@
...
+        when {
+            workflow {
+                """
+                input[0] = channel.of( [ [ id:'test' ], // meta map
+                    file('${moduleDir}/tests/data/test.fasta.gz', checkIfExists: true),
+                    file('${moduleDir}/tests/data/test.gff.gz', checkIfExists: true),
+                    file('${moduleDir}/tests/data/test.faa.gz', checkIfExists: true),
+                    file('${moduleDir}/tests/data/test_ips.tsv.gz', checkIfExists: true),
+                ])
+                input[1] = Channel.value(null)      // ch_antismash_db
```
vagkaratzas: There are leftover `Channel`s, but I wont block you for this. Just FYI

subworkflows/ebi-metagenomics/bgc_annotation/tests/main.nf.test
```diff
@@ -0,0 +1,314 @@
...
+                "gecco/run",
+                "gecco/convert",
+            ]
+        )
+        nfcoreLink("${launchDir}/library/", "${baseDir}/modules/")
+    }
+
+    test("positive test - contigs, gff, proteins, ips - sanntis") {
+
+        tag "sanntis"
```
vagkaratzas, line 37: Curious, what happens when you tag this here?

subworkflows/ebi-metagenomics/bgc_annotation/tests/main.nf.test.snap
```diff
@@ -0,0 +1,190 @@
...
+                    },
+                    {
+                        "start": 234257,
+                        "end": 330050,
+                        "label": "NRP_Polyketide"
+                    }
+                ],
+                "bgc_tools": [
+                    [
+                        "sanntis"
```
vagkaratzas, line 116: Is this the expected output?

vagkaratzas (approved):
Looks good..just left some comments but not blocking

subworkflows/ebi-metagenomics/bgc_annotation/tests/nextflow.config
Ales-ibt: done

subworkflows/ebi-metagenomics/bgc_annotation/tests/main.nf.test
Ales-ibt: I am using those tags to call the tests independently with nf-test

subworkflows/ebi-metagenomics/bgc_annotation/tests/main.nf.test.snap
Ales-ibt: Yes it is :D

subworkflows/ebi-metagenomics/bgc_annotation/tests/main.nf.test
Ales-ibt: Fixed

KateSakharova (approved):

PR #151, opened 2026-04-16
Removing hps path

jmattock5:
I was checking for any paths to our space in envident and found this one. Have tried running the test locally after removing this string and it passes


mberacochea (approved):
🥇 thanks Jenny

PR #152, opened 2026-04-21
Add bbmap reformat and reformat_standardise 

timrozday-mgnify:
Both modules (`bbmap/reformat` and `bbmap/reformat_standardise`) have been taken from raw-read-analysis-pipeline (RawR). They are now needed in the Ampicon pipeline so they should both use shared modules in this repo.

These BBMap operations are useful as an optional first stage of a pipeline, especially with non-ENA data, e.g. they will de-interleave interleaved paired end data and normalise headers

The `bbmap/reformat` module is new and the `bbmap/reformat_standardise` module is an update of an existing module (although it was only used in RawR and had since been replaced by a modified local version). I have made slight changes to the modules compared to the local RawR pipeline, and have added module tests.

In the updated `bbmap/reformat_standardise` module I have set `spaceslash=f` and `verifypaired=f` because `verifypaired=t` is time-consuming, and the space inserted by `spaceslash=t` was causing downstream errors.


modules/ebi-metagenomics/bbmap/reformat/tests/main.nf.test
```diff
@@ -0,0 +1,72 @@
...
+    // so output headers should be exactly @ERR4334351.N with no trailing tokens.
+
+    test("single-end fastq") {
+
+        when {
+            process {
+                """
+                input[0] = [
+                    [ id: 'test', single_end: true ],
+                    file("${moduleDir}/tests/data/test-reads/ERR4334351_test_1.fastq.gz", checkIfExists: true)
```
KateSakharova, line 26: You can use datasets from nf-core. Check https://github.com/nf-core/modules/blob/master/modules/nf-core/bbmap/align/tests/main.nf.test#L31

modules/ebi-metagenomics/bbmap/reformat/environment.yml
```diff
@@ -0,0 +1,7 @@
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/environment-schema.json
+channels:
+  - conda-forge
+  - bioconda
+dependencies:
+  - "bioconda::bbmap=39.33"
```
KateSakharova, line 7: there should be pigz as well
like https://github.com/nf-core/modules/blob/master/modules/nf-core/bbmap/bbmerge/environment.yml#L8

modules/ebi-metagenomics/bbmap/reformat/main.nf
```diff
@@ -0,0 +1,57 @@
...
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/5a/5aae5977ff9de3e01ff962dc495bfa23f4304c676446b5fdf2de5c7edfa2dc4e/data' :
+        'community.wave.seqera.io/library/bbmap_pigz:07416fe99b090fa9' }"
+
+    input:
+    tuple val(meta), path(reads)
+    val(out_fmt)
+
+    output:
+    tuple val(meta), path("*_reformated.${out_fmt}")       , emit: reformated
```
KateSakharova: align emits :)

modules/ebi-metagenomics/bbmap/reformat/main.nf
```diff
@@ -0,0 +1,57 @@
...
+    path  "versions.yml"                                      , emit: versions
```
KateSakharova: version can be topic
https://github.com/nf-core/modules/blob/master/modules/nf-core/bbmap/bbmerge/main.nf#L18

modules/ebi-metagenomics/bbmap/reformat/main.nf
```diff
@@ -0,0 +1,57 @@
...
+    path  "*.log"                                             , emit: log
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def args   = task.ext.args ?: ''
+    prefix = task.ext.prefix ?: "${meta.id}"
```
KateSakharova: add `def` to prefix and in_reads and out_reads

modules/ebi-metagenomics/bbmap/reformat/main.nf
```diff
@@ -0,0 +1,57 @@
...
+    in_reads  = "in=${reads}"
+    out_reads = "out=${prefix}_reformated.${out_fmt}"
+
+    """
+    maxmem=\$(echo \"$task.memory\"| sed 's/ GB/g/g')
+    reformat.sh \\
+        -Xmx\$maxmem \\
+        $in_reads \\
+        $out_reads \\
+        threads=${task.cpus} \\
```
KateSakharova, line 34: If you use threads then process can't be `single` (line 3). It should be medium probably...

modules/ebi-metagenomics/bbmap/reformat/main.nf
```diff
@@ -0,0 +1,57 @@
...
+        uniquenames=t \\
+        trimreaddescription=t \\
```
KateSakharova: I think those should go into args if you try to make a universal module. You can add config to that file like https://github.com/EBI-Metagenomics/nf-modules/blob/main/subworkflows/ebi-metagenomics/amr_annotation/tests/nextflow.config

modules/ebi-metagenomics/bbmap/reformat/meta.yml
```diff
@@ -0,0 +1,60 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/meta-schema.json
+name: "bbmap_reformat"
+description: "Reformat and convert sequencing reads using BBMap's reformat.sh tool. Handles single-end and interleaved FASTQ/FASTA files, adding unique names and trimming read descriptions."
```
KateSakharova: If that tool also handles interleaved fq then it also should be something like here https://github.com/nf-core/modules/blob/master/modules/nf-core/bbmap/bbmerge/main.nf#L11 and https://github.com/nf-core/modules/blob/master/modules/nf-core/bbmap/bbmerge/main.nf#L26

modules/ebi-metagenomics/bbmap/reformat_standardise/main.nf
```diff
@@ -14,7 +14,8 @@ process BBMAP_REFORMAT_STANDARDISE {
     output:
     tuple val(meta), path("*_reformated.${out_fmt}")       , emit: reformated
     tuple val(meta), path("${prefix}_singleton.${out_fmt}"), optional: true, emit: singleton
-    path  "versions.yml"                                   , emit: versions
+    path  "versions.yml"                                      , emit: versions
```
KateSakharova: align emits :)

modules/ebi-metagenomics/bbmap/reformat_standardise/main.nf
```diff
@@ -26,19 +27,21 @@ process BBMAP_REFORMAT_STANDARDISE {
...
     reformat.sh \\
         -Xmx\$maxmem \\
         $in_reads \\
         $out_reads \\
         $interleaved_cmd \\
         $paired_cmd \\
         threads=${task.cpus} \\
-        ${args}
+        allowidenticalnames=t \\
+        trimreaddescription=t \\
```
KateSakharova: the same here

KateSakharova (changes requested):
Thank you for fixes! I have left some comments mostly about `reformat`

modules/ebi-metagenomics/bbmap/reformat/main.nf
timrozday-mgnify: Yes that would be better, I've made changes

modules/ebi-metagenomics/bbmap/reformat_standardise/main.nf
timrozday-mgnify: I've now changed this to something a little different. Because some other arguments are hardcoded based on conditions from `meta` (`single_end` and `interleaved`), I have args, paired_args and interleaved_args all modifiable from config (`task.ext`), for maximum flexibility

modules/ebi-metagenomics/bbmap/reformat/meta.yml
timrozday-mgnify: My mistake, I think that was a copy-paste from bbmap_reformat_standardise. It doesn't handle paired-end or interleaved, only single-end or merged. I've ammended the description now

modules/ebi-metagenomics/bbmap/reformat/tests/main.nf.test
timrozday-mgnify: Do I have to?

modules/ebi-metagenomics/bbmap/reformat/main.nf
```diff
@@ -0,0 +1,55 @@
...
+    maxmem=\$(echo \"$task.memory\"| sed 's/ GB/g/g')
+    reformat.sh \\
+        -Xmx\$maxmem \\
+        $in_reads \\
+        $out_reads \\
+        threads=${task.cpus} \\
+        ${args} \\
+        &> ${prefix}.reformat.sh.log
+
+    cat <<-END_VERSIONS > versions.yml
```
KateSakharova: If you add topic like I sent you example from nf-core than you can skip that block and in stub as well

modules/ebi-metagenomics/bbmap/reformat_standardise/main.nf
```diff
@@ -14,31 +14,33 @@ process BBMAP_REFORMAT_STANDARDISE {
...
+    def args = task.ext.args ?: ''
+    def interleaved_args = task.ext.interleaved_args ?: ''
+    def paired_args = task.ext.paired_args ?: ''
     prefix = task.ext.prefix ?: "${meta.id}"
-    single_file = (reads instanceof Collection) ? (reads.size() == 1) : true
-    in_reads  = single_file ? "in=${reads[0]}" : "in=${reads[0]} in2=${reads[1]}"
-    out_reads = meta.single_end ? "out=${prefix}_reformated.${out_fmt}" : "out=${prefix}_1_reformated.${out_fmt} out2=${prefix}_2_reformated.${out_fmt} outs=${prefix}_singleton.${out_fmt}"
-    interleaved_cmd = meta.interleaved ? "int=t verifyinterleaved=t" : ""
-    paired_cmd = meta.single_end ? "" : "addslash=t spaceslash=t verifypaired=t"
+    def in_reads = (meta.single_end || meta.interleaved) ? "in=${reads[0]}" : "in=${reads[0]} in2=${reads[1]}"
```
KateSakharova: I would not use meta.interleaved because it is absolutely hidden and not obvious for user that this argument is important. That is why for example fastp module has so many inputs. It is not a problem to call process PROCESS(input, [], []) in case you do not need those. But it is very annoying to modify meta before every run of that PROCESS

KateSakharova (changes requested):

modules/ebi-metagenomics/bbmap/reformat_standardise/main.nf
timrozday-mgnify: OK, I've put it as a separate input

modules/ebi-metagenomics/bbmap/reformat/main.nf
timrozday-mgnify: Oh I see, and this is preferred? OK I've replaced it

KateSakharova (approved):

PR #155, opened 2026-07-10
Bump the version of GP -- support for gz IPS TSV now

mberacochea:
I've patched GP to support .gz interproscan tsv now. This will help with the storage bloat.

GP changes -> https://github.com/EBI-Metagenomics/genome-properties/releases/tag/v2.0.2


vagkaratzas (approved):
Awesome

modules/ebi-metagenomics/genomeproperties/tests/main.nf.test.snap
```diff
@@ -26,7 +26,7 @@
                             "id": "test",
                             "single_end": false
                         },
-                        "test_gp.txt:md5,d41d8cd98f00b204e9800998ecf8427e"
+                        "test_gp.txt.gz:md5,d41d8cd98f00b204e9800998ecf8427e"
```
chrisAta, line 29: I might be way off, but this is the md5sum for an empty file, but aren't compressed files never actually empty cause of header metadata etc?

modules/ebi-metagenomics/genomeproperties/tests/main.nf.test.snap
```diff
@@ -35,7 +35,7 @@
                             "id": "test",
                             "single_end": false
                         },
-                        "test_gp.json:md5,d41d8cd98f00b204e9800998ecf8427e"
+                        "test_gp.json.gz:md5,d41d8cd98f00b204e9800998ecf8427e"
```
chrisAta, line 38: same here

chrisAta (approved):

modules/ebi-metagenomics/genomeproperties/tests/main.nf.test.snap
mberacochea: These are the stubs, which were empty before too.

